PR #48, opened 2023-02-14
New: Monitor performers via Lists

Qstick:
#### Database Migration
NO

#### Description
Lists to monitor performers

#### Todos
- [ ] Tests

### DIFF
--- a/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
+++ b/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
@@ -22,6 +22,12 @@ import styles from './EditImportListModalContent.css';
 
 function EditImportListModalContent(props) {
 
+  const monitorOptions = [
+    { key: 'none', value: 'None' },
+    { key: 'specificEpisode', value: 'Specific Episode' },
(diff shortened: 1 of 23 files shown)


salesawagnerpe:
any news?

guim31:
That would be awesome ;)

i-am-at0m:
Um I think this might be broken it has added so many sites and it's monitoring every episode on every single one

thezak48:
> Um I think this might be broken it has added so many sites and it's monitoring every episode on every single one

You need to set monitor to specific episodes
image

guim31:
Dumb question sorry : how do I install this version of whisparr ?

thezak48:
> Dumb question sorry : how do I install this version of whisparr ?

it is in the nightly builds

guim31:
> > Dumb question sorry : how do I install this version of whisparr ?
> 
> it is in the nightly builds

Here is the source I use in my Unraid setup : cr.hotio.dev/hotio/whisparr:nightly

Should I be more precise on which nighlty build I want ?

thezak48:
as long as you are running the latest build then you will have it

PR #121, opened 2023-10-11
New on-theme flavor text

qeasd42:
#### Database Migration
NO

#### Description
The nzbdrone flavor text wasn't doing it for me.  This should be a healthy amount of fresh content.  Quotes pulled from popular novels, films, adult-films, stand-up jokes, and songs.  It would appear none of this lives in the localizations so I felt safe using puns/quotes from English only.

Admittedly some of these quotes are extremely long and might break the UI.  Then again, I heard the UI likes it that way,

#### Todos
- [x] Tests

* I will never be satiated


Gykes:
I like the idea but some of the quotes are rather.... vulgar. I know what this app is and all but a lot of users prefer to be as discreet as possible, this may make it less so.

PR #150, opened 2024-01-24
Fix: Performer related fixes and additions

Tom-Estes:
Database Migration
NO

Description
- When adding a performer, the option to search for scenes has been fixed.
- When a Site has 2 or more releases on the same date, the indexer will include all scenes with that date. By adding the performer to the indexer search, it narrows down the amount of scenes returned.
- Added performer naming tokens (male & female, female only).
- Added performers option to the scene row in both performer view and studio view.

### DIFF
--- a/frontend/src/AddMovie/AddNewMovie/AddNewMovie/AddNewMovieSearchResult.js
+++ b/frontend/src/AddMovie/AddNewMovie/AddNewMovie/AddNewMovieSearchResult.js
@@ -97,12 +97,13 @@ class AddNewMovieSearchResult extends Component {
     if (itemType === 'scene') {
       ImageItem = ScenePoster;
       posterWidth = 300;
-      posterHeight = 167;
+      posterHeight = 169;
     }
 
     const elementStyle = {
       width: `${posterWidth}px`,
-      height: `${posterHeight}px`
+      height: `${posterHeight}px`,
+      objectFit: 'contain'
     };
 
     return (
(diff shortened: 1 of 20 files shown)


src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -116,7 +117,7 @@ public virtual IndexerPageableRequestChain GetSearchRequests(SceneSearchCriteria
                     pageableRequests.Add(GetPagedRequests(MaxPages,
                         Settings.Categories,
                         "search",
-                        $"&q={NewsnabifyTitle(queryTitle)}+{releaseDate}"));
+                        $"&q={NewsnabifyTitle(queryTitle)}+{releaseDate}+{performer}"));
```
Qstick: Does this makes results too restrictive? What if the performer is not in the release name

Qstick (commented):
Nice changes once again.... Just one comment about the search change

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
Tom-Estes: The reason I added it was I ran a search without the performer using 24.01.20 and it returned 2 pages with dates ranging from 24.01.20-20.01.24
I can add a check that if no results are returned and run the search again without the performer for each query. Or add additional criteria to the queryTitles like title+performer, site+title, site+performer. I don't know if the indexer will see it as a ddos attack.

Qstick: Yea, let's add a second tier for that. The tier logic will handle that automatically with rate limits

Tom-Estes: Gonna need to do some more testing. I searched for a known release without the performer in the title and we still rejected the release with the original query. Did you want to exclude this commit from the merge for now?

Qstick: Yea, that works until we get things sorted

Qstick (approved):

PR #151, opened 2024-01-25
Fix: Search on add for scene when Studio Network is in the filename

sampulsar:
#### Database Migration
NO

#### Description
Allow the Studio Network token to be used when the studio is not currently in the database by passing the value from the search results. Also null check the studio from the database as it can be null.


sampulsar:
fails on database

PR #154, opened 2024-02-09
Fix: renaming scenes movies

Tom-Estes:
#### Database Migration
NO

#### Description
- Cast mapping was defaulting to male resulting in not filtering by gender when using performerFemale naming token
- OCD performer ordering by first name
- Use StashId in movie cast to enable using the same performer id in movies, scenes and performer page
- Previously had to enable both movie renaming and scene renaming to rename scenes


Qstick:
Looks good, can you squash these back into the original 4 commits, then neame the commits using `Fixed` instead of `Fix`. This ensures they show up in the app change log

PR #157, opened 2024-02-16
Fixed: date mismatch for episode air date

### LINKED ISSUE #109: Date mismatch, probably due to timezone

### Current Behavior

Whisparr is one day off sometimes from tpdb. Probably due to timezone. It would be better to only use date instead of date and time (if this is the problem)

E.g. tpdb:
Screenshot_2023-08-28-16-09-39-44_40deb401b9ffe8e1df2f1cc5ba480b12~2

Whisparr:
Screenshot_2023-08-28-16-08-41-31_40deb401b9ffe8e1df2f1cc5ba480b12~2

### Expected Behavior

Use the date in tpdb, now it seems to do a relative date to the tpdb date...

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: unraid docker
- Whisparr: 2.0.0.250, nightly-51f4a29 by hotio
- Docker Install: yes
- Using Reverse Proxy:no
- Browser: chrome
```

Main

### Trace Logs?

N/a

### Anything else?

_No response_

CyyberFroggy:
#### Database Migration
NO

#### Description
The old way of parsing left the DateTime with DateTimeKind.Unspecified. This resulted in a conversion to UTC when written to the database and thus a different date for everybody in a timezone ahead of UTC (for example UTC+1). When we parse the DateTime with DateTimeStyles.AssumeUniversal it sets the kind to DateTimeKind.Utc and no conversions happen.

To get the correct date for existing scenes a refresh of the site in Whisparr is enough.

#### Todos
- [X] Tests

* Fixes #109

### DIFF
--- a/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
+++ b/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
@@ -228,7 +228,7 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
             episode.Runtime = oracleEpisode.Duration.GetValueOrDefault();
 
             episode.AirDate = oracleEpisode.ReleaseDate;
-            episode.AirDateUtc = DateTime.Parse(episode.AirDate, DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AdjustToUniversal);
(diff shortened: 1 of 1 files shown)


Gykes:
Looks good, `AssumeUniversal` is better as it was more than likely causing local time shifts without it.

PR #172, opened 2024-03-13
Fixed: V3 Filter scenes by studio(s)

plz12345:
#### Database Migration
NO

#### Description
Eros had a small bug in scene custom filters where when creating one based on one or more studios, it would always return the `All scenes hidden due to filter` message.  I dug in, and found that the filter being saved to the DB was using `studio` instead of `studioTitle`, where the latter is the proper column on the Movie table.  This change only fixes it for Scenes, which was definitely broken.  I don't know if the same holds true for Movies as I don't have any with a non-null studio in this build.  Before/after custom filter posted [on Discord](https://discord.com/channels/957516730727563285/957516731843215360/1217335696709386320).

#### Todos
- [x] Tests

### DIFF
--- a/frontend/src/Store/Actions/sceneIndexActions.js
+++ b/frontend/src/Store/Actions/sceneIndexActions.js
@@ -193,7 +193,7 @@ export const defaultState = {
       valueType: filterBuilderValueTypes.RELEASE_STATUS
     },
     {
-      name: 'studio',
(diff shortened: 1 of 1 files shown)


plz12345:
Revised

Tom-Estes (approved):

PR #203, opened 2024-03-25
Fixed: Omit special characters when using filters that are not expecting them

### LINKED ISSUE #178: V3: Filter On Manual Import

Once the list of scenes gets long enough. Searching by name causes significant delay and lag between typing each character and the list refreshing. Sometimes 3-4 seconds while the list filters itself.

Have the ability to narrow the list initially by selecting a studio or year to help prevent the search lag.

None that would not require a major rework.

### Anything else?

_No response_

plz12345:
#### Database Migration
NO

#### Description
Previously, the SelectMovieModalContent and MovieSearchInput compoents both allows for special characters, which could pollute, and in the case of the former, negate desired filtering results.  This change employs string redaction to remove the special characters before they would execute a filter.

I will note that the MovieSearchInput (top bar) filtering is happening in realtime, so it has the effect of ignoring special characters from the input, even if pasted.  The SeelctMovieModalContent simply ignores them silently.  these two mechanisms are implemented in completely different ways, and I didn't want to rewire one of them for an edge case.  I did not employ a trim() on either as it would cause inconsistent handling of trailing spaces and in testing, actually prevented spaces on the top bar filter.

Also, I made an attempt to improve filtering performance on larger (100K+) libraries in the import modal, pending feedback from #178 .

image

#### Todos
- [x] Tests

* Fixes #178

### DIFF
--- a/frontend/src/InteractiveImport/Movie/SelectMovieModalContent.tsx
+++ b/frontend/src/InteractiveImport/Movie/SelectMovieModalContent.tsx
@@ -172,9 +172,9 @@ function SelectMovieModalContent(props: SelectMovieModalContentProps) {
 
     return sorted.filter(
       (item) =>
-        item.title.toLowerCase().includes(filter.toLowerCase()) ||
-        item.tmdbId.toString().includes(filter) ||
-        item.imdbId?.includes(filter)
+        item.sortTitle.includes(
+          filter.toLowerCase().replace(/[^a-zA-Z ]/g, '')
+        ) || item.tmdbId.toString() === filter
     );
   }, [allMovies, filter]);
 
--- a/frontend/src/Components/Page/Header/MovieSearchInput.js
+++ b/frontend/src/Components/Page/Header/MovieSearchInput.js
@@ -134,7 +134,9 @@ class MovieSearchInput extends Component {
       return;
     }
 
-    this.setState({ value: newValue });
+    this.setState({
+      value: newValue.replace(/[^a-zA-Z ]/g, '')
+    });
   };
 
   onKeyDown = (event) => {


frontend/src/InteractiveImport/Movie/SelectMovieModalContent.tsx
```diff
@@ -172,9 +172,11 @@ function SelectMovieModalContent(props: SelectMovieModalContentProps) {
...
     return sorted.filter(
       (item) =>
-        item.title.toLowerCase().includes(filter.toLowerCase()) ||
-        item.tmdbId.toString().includes(filter) ||
-        item.imdbId?.includes(filter)
+        item.sortTitle.includes(
+          filter.toLowerCase().replace(/[^a-zA-Z ]/g, '')
+        ) ||
+        item.tmdbId.toString() === filter ||
+        item.imdbId === filter
```
Tom-Estes: Is imdbId still being used? I believe you removed the link in #191

plz12345: Sorry, must have spaced on that.  Will revise it this weekend and push.  IIRC, I had a hard time completely removing it, and undid that just to get this fix submitted, since the fix isn't technically related to removing the IMDB tech debt.

Tom-Estes: No worries, just an observation. Wanted to make sure, we can clean up as we go.

plz12345: Removed IMDB from here.  I did not go on a mission to find any and all traces in other files, but at least it's gone from the ones I've touched for this fix.

Tom-Estes (approved):

PR #215, opened 2024-04-13
Fixed: V3 No results in Scene Search if studio name contains a space

### LINKED ISSUE #213: V3 unable to handle search on studios with spaces in names

### Current Behavior

When trying to use search (both auto and interactive) for a studio that has a space in the name, v3 fails to effectively pull back releases.

**Examples**  (Whisparr currently does these when using Search)
`Pure Taboo 24.04.09` - no results
`Parental Guardian Angel 24.04.09` - no results

This was [mentioned](https://discord.com/channels/957516730727563285/957516731654488087/1227633157000532018) the other day on Discord, but I didn't get a chance to test at the time to confirm.  This is that confirmation.

The as-is queries fail to pull any results, however `puretaboo 24.04.09` pulls releases from multiple indexers.  RSS-pulled titles do not seem to hit this limitation, so I imagine it's more robust in title processing there vs. what is used in searches.

### Expected Behavior

Either remove the spaces in studio names for the query, or use the Studio's `cleantitle` for the Studio name being passed into the search, since that column is already formatted well for this.

**Example**
`puretaboo 24.04.09` - expected results are found

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Linux
- Whisparr: 3
- Docker Install: yes
- Using Reverse Proxy: no
- Browser: Firefox
- Database: SQLite
```

Main

### Trace Logs?

Available if needed.  I was using Prowlarr history to validate Whisparr's queries vs. desired format

### Anything else?

_No response_

plz12345:
#### Database Migration
NO

#### Description
When searching for a Scene using either Interactive or Automatic search, there would be no results if the Studio the scene belongs to had a space in it.  This was fixed in V2 by also searching `TitleSlug` in the same .NET class, but the code in V3 was a bit different.  I used the same strategy, though, by adding an additional search string if the Studio has a space in the name.

This was only an issue with user-triggered searches, not RSS grabs.  The release title parser is more robust in the RSS pulls.  I wanted to get this fix in because we still have quite a bit of lag in the SkyHook sync to StashDB, to the point where RSS may roll past desired releases before Whisparr can add them.  This fix allows the user to manually add and search a specific scene more reliably than without the de-spaced Studio title.

#### Todos
- [X] Tests

* Fixes #213

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
@@ -71,6 +71,11 @@ namespace NzbDrone.Core.IndexerSearch
                     sceneSearchSpec.SceneTitles.Add(sceneSearchSpec.SiteTitle);
                 }
 
+                if (sceneSearchSpec.SiteTitle.Contains(' ', StringComparison.Ordinal))
+                {
+                    sceneSearchSpec.SceneTitles.Add(sceneSearchSpec.SiteTitle.Replace(" ", "", StringComparison.Ordinal));
+                }
+
                 decisions = await Dispatch(indexer => indexer.Fetch(sceneSearchSpec), sceneSearchSpec);
             }
 


pythongod (approved):

nod44 (approved):
Seems like a reasonable change.

nod44:
@Qstick Do you approve PRs on this repo?  Can you take a look at some of these PRs, or how can we help?

sampulsar:
Should also clean the studio title with  CleanMovieTitle() so that it is closer to v2. This cleans any additional characters such as singe quotes, etc.

plz12345:
> Should also clean the studio title with CleanMovieTitle() so that it is closer to v2. This cleans any additional characters such as singe quotes, etc.

Can you merge after?  It's been sitting for 6 weeks at this point.

Tom-Estes (approved):

PR #216, opened 2024-04-15
V2: Test ByAirDateFixture

sampulsar:
#### Database Migration
NO

#### Description
Issue raised in discord that an import is not matching when the title is part of a file name and correctly parsed. Used the unit tests to verify any issues.

- Test with female performers (should_get_episode_when_two_regular_episodes_share_the_same_air_date_and_performer_and_part_is_provided then fails)
- Create new test for when a performers name is within the title
- Refactor FindOneByAirDate so it can match as various levels (Exact Title match being the highest) (Tests now pass)
- SixLabors.ImageSharp required an upgrade as it has a security issue now.

#### Todos
- [x] Tests


Tom-Estes (approved):

rahspx:
This had some unintended issue. For properly renamed episode without the performer, 2 or more same day episode, it switched them around not getting the right episode title. It suggested renaming to the other episode. Even after assigning and correcting manually to the proper episode, during the rescan/update it again updates to the wrong title. There is no exception but suggest the wrong episode during rename preview.

PR #233, opened 2024-08-11
Fix: Scene page breaking after deleting scenes

kaldigo:
#### Database Migration
NO

#### Description
After deleting a lot of scenes opening the scene page would error and this fixes that.


Tom-Estes:
Can you squash these into1 commit, then rename the commits using Fixed instead of Fix. This ensures they show up in the app change log

sampulsar:
Completed in #290

PR #257, opened 2024-09-28
Fixed incorrect error message

artsmarupid:
The error message is displayed when not applicable, and the error message does not include the correct appName. The fix is cherry-picked from https://github.com/Sonarr/Sonarr/pull/6420


Gykes:
I merged the Cherry pick from Sonarr

PR #273, opened 2024-10-18
Eros - Find Best Match

sampulsar:
#### Database Migration
NO

#### Description
Normalise the ReleaseTokens value so that it can match more often, & to and, remove any space substitutions from special characters.
Search all of the studios in case a duplicate match exists across studios potential.
No longer accept if a single item is return for a studio on a date.
Rank the matches and use the best match if there is multiple, MovieParseMatchType order.

#### Todos
- [x] Tests

* Fixes #XXXX


Tom-Estes (changes requested):
Can you update this branch to the latest and see if the checks pass after?

sampulsar:
@Tom-Estes 

Could not update the branch, so I created an new one and created #277. It is slightly different and is in the comments for the PR.

PR #292, opened 2024-11-10
New: Parse new format, [studio] title [dd.mm.yyyy, that is also used as a search term for the indexers 

sampulsar:
#### Database Migration
NO

#### Description
Add additional parser for a new format returned in a search.
Parse [studio] title [dd.mm.yyyy

Remove searching on just Studio Name as it returns too many irrelevant results.

Some additional error handlers

#### Todos
- [ ] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/ParseMovieTitleFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/ParseMovieTitleFixture.cs
@@ -7,6 +7,9 @@ namespace NzbDrone.Core.Test.ParserTests
     [TestFixture]
     public class ParseMovieTitleFixture : CoreTest
     {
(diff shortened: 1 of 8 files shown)


Tom-Estes (changes requested):
Can you separate the areas of concern into their own commits? They can be in the same PR or branched out if you want, but it seems jumbled based on the title.

Also, you're removing Site + title, but wasn't it just added in #276? I believe it was added to help the searching in torrents. Maybe find a way to move it to torrent only searches.

sampulsar:
@Tom-Estes 

Split into the torrent searching and parsing, and then the additional error handling and logging.

The removal from #276 was just searching on the studio name that resulted in way too many false results.

sceneSearchSpec.SceneTitles.Add(sceneSearchSpec.SiteTitle); eg Brazzers Exxtra

The loop goes through and then adds Brazzers Exxtra 24.11.11, Brazzers Exxtra 11-11-2024, Brazzers Exxtra {title of scene}

Just the title was added to return more results.

sceneSearchSpec.SceneTitles.Add(sceneSearchSpec.Movie.Title); 

In most case it should get a good selection of results.

The next thing would be studio aliases, Brazzers Exxtra (Done), BrazzersExxtra (Done),  ??? (Alias on StashDB once feature is avaliable)

Tom-Estes (approved):

Issue #299, opened 2024-11-14
[v3] Scenes that should be rejected are grabbed by automatic search


gujikabeepbop:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Current Behavior

For studios releasing multiple scenes a day (BrazzersExxtra) the downloaded scene is not always correct. Sometimes it is another release of the same day.

When I do a manual search on a BrazzersExxtra scene I see the other scenes of the day are rejected releases because of the reason 'Unknown Movie' (which is a correct behaviour). But the automatic search will still sometimes grab the wrong scene. 

### Expected Behavior

Automatic search should never grab scenes that are rejected because of 'Unknown Movie'

### Environment

```markdown
whisparr v3
```

### What branch are you running?

Main

### Trace Logs?

can provide if needed.

sampulsar:
This is known and the logic needs to be removed, or toggled off within Whisparr

https://github.com/Whisparr/Whisparr/blob/eros/src/NzbDrone.Core/Movies/MovieService.cs#L412

gujikabeepbop:
The config knob doesn't work in the UI. I think something is missing in src/Whisparr.Api.V3/Config/HostConfigResource.cs.

PR #307, opened 2024-11-18
New: Cache API to memory Movie and Performer Resources

sampulsar:
#### Database Migration
NO

#### Description

Cache the built versions of the resources for Movies\Scenes and Performers. This is done with a read through cache.

Use a lock to reduce the load on the server by ensuring that the items are built a single time.

Discard the cached item if deleted or updated.

Extract Performers Scenes from cache to reduce the load on the database doing a text search.

Setting for to toggle and correct #299 

The default is the current behaviour and both need to be toggled on.

image

#### Todos
- [x] Tests

* Fixes #XXXX


plz12345:
Will this negate the expensive API call to bulk fetch all movies/scenes when open the app in a new tab?

sampulsar:
@plz12345 

This will reduce the processing that the server needs to do to respond to the request to return data from the movie/scene API. The client still needs to receive and parse the data. The only way to make the client even faster is to store a copy of the data on the client using indexeddb.

A change to indexeddb would be a large change to the client.

sampulsar:
No longer think this is required as it can be corrected by fixing the underlying issue, and normalise the data that the client has #318

PR #310, opened 2024-11-22
New: Add Studio Performer Exclusions

### LINKED ISSUE #188: V3: Blocklist

Unable to blocklist studios or scenes.

Would like to blocklist so that they don't get added and take up space.

None

### Anything else?

Requested per Plz from another issue.

sampulsar:
#### Database Migration
YES - ImportExclusions Add Type

#### Description
Allow an Exclusion to be added for a Studio or Performer. This will then add a Scene exclusion.

The Exclusions not have a type and can be added via /settings/importlists

image

#### Todos
- [x] Tests

* Fixes #188


plz12345:
Could this work without adding the scene exclusions, too?  That list isn't optimized for large quantities, in both page loading as well as lack of bulk management.  I'd worry it could eventually start to be a performance issue when viewing the Import Lists page.  I know the trade-off would be that Whisparr would evaluate the same scenes for studio exclusions on every list sync, but it'd be doing that either way against the exclusions list itself.

sampulsar:
@plz12345 

The exclusions are on the scenes id as this reduces the load on the skyhook server, as studios and performers only list the scene ids. This server is hosted and the hosting costs money, so retrieving a performers list of scenes and constantly retrieving some scenes daily, just to discard the data will increase the cost of this server.

If the majority of scenes are excluded then likely an import list should be used.

plz12345:
> @plz12345 
> 
> 
> 
> The exclusions are on the scenes id as this reduces the load on the skyhook server, as studios and performers only list the scene ids. This server is hosted and the hosting costs money, so retrieving a performers list of scenes and constantly retrieving some scenes daily, just to discard the data will increase the cost of this server.
> 
> 
> 
> If the majority of scenes are excluded then likely an import list should be used.

I'm using import lists exclusively.  I'm going to use tag exclusions to avoid stuff like VR that I don't care about.  Maybe I will take a crack at making the exclusion list view more efficient to scale. No worries.

plz12345:
I refactored the import exclusion table a bit, but will wait for this PR to merge and rebase.  The only C# code I touched is a translation, which @sampulsar probably already covered.

PR #342, opened 2025-02-09
Fix history retention error message

meerumschlungen:
#### Database Migration
NO

#### Description
This is a quick fix for the following situation:
If the SABnzbd config property `history_retention` is an empty string, the retention will be infinite. Whisparr has a false positive error message in this case, stating that the download client would probably remove downloads before "Sonarr" [sic] could grab them.

image

#### Todos
- ~[ ] Tests~
- ~[ ] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)~
- ~[ ] [Wiki Updates](https://wiki.servarr.com)~


sampulsar (approved):

meerumschlungen:
PR unnecessary, original issue solved by #389.

PR #393, opened 2025-03-09
Changed studio link in SceneIndexRow from white to defaultColor

adtdev20:
#### Database Migration
NO

#### Description
Light mode rendering causes the Studio Name and Link to be pretty much invisible when using the Scene Table view. The existing css has the color set to "White". Updated it to defaultColor so it will adjust with changing between light and dark mode.

**Before**
before

**After**
after

**Dark Mode continues to work as before**
dorkmode

### DIFF
--- a/frontend/src/Scene/Index/Table/SceneIndexRow.css
+++ b/frontend/src/Scene/Index/Table/SceneIndexRow.css
@@ -35,7 +35,7 @@
 .studio>a:visited,
 .studio>a:hover,
 .studio>a:active {
-  color: var(--white);
+  color: var(--defaultColor);
 }
 
 .originalLanguage,


Gykes:
Although it's not required. I like to add a before and after of UI changes so that regular users can see the differences. 

example: https://github.com/Whisparr/Whisparr/pull/392

plz12345:
Heh, my fault for not testing in light mode.  Dark mode is too superior.

adtdev20:
> Although it's not required. I like to add a before and after of UI changes so that regular users can see the differences.
> 
> example: #392

Updated pr comment with screenshots, good call

Tom-Estes (approved):

PR #396, opened 2025-03-11
Adding TPDBID to Site Page

Gykes:
Adding in another label in the site details section to show the Site ID. Users find it complicated to either click the link and go to a whole new site or search TPDB manually with the name to find the ID.

As you can see by the preview it adds the Site ID next to the Network Label.

before
vs
after

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.js
+++ b/frontend/src/Series/Details/SeriesDetails.js
@@ -527,6 +527,24 @@ class SeriesDetails extends Component {
                       </Label>
                   }
 
+                  {
+                    !!tvdbId && (
+                      <Label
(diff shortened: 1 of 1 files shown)


thezak48:
Why not have the site id show up as a tooltip on hover?

Gykes:
> Why not have the site id show up as a tooltip on hover?

Hover on what? I could have it on hover of the TPDB id but that just seemed redundant. 

Based on a convo with a user In disc I specifically made it its own label because the user needs to know what the id for several sites are. Currently they have to click on the link url and find the id on TPDB and this is a lot of clicks. This way the user can just open the site and have the id shown to them.

sampulsar (approved):

PR #398, opened 2025-03-12
TPDB ID Tooltip Hover Table View

Gykes:
Adds the ability for users to hover over site hyperlinks in Table view on sites tabs and see TPDB ID. 

Discord user wanted to be able to get all the IDs from their sites a little faster. Doing it this way reduces the visual clutter but allows the users to speed up the process.

Example: 

example


Gykes:
Bunch of commit errors, just going to close and repush

PR #404, opened 2025-03-15
Fix: Performers Delete Button was Accidentally Moved

Gykes:
When creating the expand/collapse all for performers I accidently put it above the delete button causing the delete button to move to the far right.


Gykes:
Closing due to GH errors.... will Re PR

PR #405, opened 2025-03-15
Fix: Reposition Performer Delete Button On Left Side

Gykes:
When creating the expand all button I accidently placed it above the delete button causing the delete button to be forced to the far right as it was inside the toolbar section.


plz12345:
This looks good, but you'll need to fix the linting and squash the commits before merge.

image

Tom-Estes (approved):

PR #413, opened 2025-03-18
New: added Studio cleantitleslug

Tom-Estes:
#### Database Migration
NO

#### Description
Removes spaces in studio name


gujikabeepbop:
Shouldn't this also remove single quotes?

PR #422, opened 2025-03-22
NEW: Add Monitor/Unmonitor All To Studio Page

### LINKED ISSUE #181: V3: Batch Monitoring and Unmonitoring

Presently it is not possible to monitor/unmonitor entire years, missing or new only as some people are used to.

Given that stash is in use and scenes are not automatically added but curated. This can cause scenes to appear much later then released.

Add the ability to at least unmonitor years of site scenes. And persist so that 'new' scenes for that year if they are found do not come in with a monitored flag.

All new scenes found outside of the 'current' latest year be unmonitored, or provide an option if new scenes should be monitored or not. Right now the only option is to search on a new scene.

### Anything else?

_No response_

Gykes:
This adds the ability to monitor/unmonitor all scenes within a year inside a studio page similar to Sonarr with series.

* Fixes #181

### DIFF
--- a/frontend/src/Studio/Details/StudioDetailsYear.js
+++ b/frontend/src/Studio/Details/StudioDetailsYear.js
@@ -11,6 +11,7 @@ import Menu from 'Components/Menu/Menu';
 import MenuButton from 'Components/Menu/MenuButton';
 import MenuContent from 'Components/Menu/MenuContent';
 import MenuItem from 'Components/Menu/MenuItem';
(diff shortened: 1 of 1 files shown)


sampulsar (approved):

sampulsar (commented):
Do you want to remove isDisabled={!studioMonitored} so that it can be used on any studio?

Gykes:
> Do you want to remove isDisabled={!studioMonitored} so that it can be used on any studio?

Good catch, only did testing on monitored studios. 

Updated

PR #419, opened 2025-03-21
V3 Fix: Hide interactive search on Performer and Studio

plz12345:
#### Database Migration
NO

#### Description
At present, Interactive Search only works when viewing a single scene, not a performer or studio.  The level of effort to fix these is heavy, so I opted to hide them until later efforts can be put into redesigning them.  I also added a missing English translation for `SearchStudio` to display as `Search Studio`.

#### Screenshot (if UI related)
The button was hidden on Performer Detail and Studio Detail.

#### Todos
- [X] Tests
- [X] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)
- [X] [Wiki Updates](https://wiki.servarr.com)

#### Issues Fixed or Closed by this PR

* Fixes #417 , #418

PR #429, opened 2025-03-23
FIX: Remove Interactive Search From Years Within Studio page

### LINKED ISSUE #427: Fix Interactive Search Within Studio Details
Interactive search within a studio errors out as it searches for the movieID rather than looping through each scene. Would need to implement something like Sonarr to allow it to loop through each scene and make individual searches.

Gykes:
Currently Interactive search within a year just does nothing. 

I am removing the front end as it won't be that useful anyway. Radarr doesn't have this functionality as it expects 1 search per studio (Movie). Sonarr has this ability to search per year but it searches for packs/siterips.

We could potentially make it work with siterips but would require significant work. 

@plz12345 Already has a PR open to remove interactive search from Studio and performer details headers 

#419

* Fixes #427

### DIFF
--- a/frontend/src/Studio/Details/StudioDetailsYear.js
+++ b/frontend/src/Studio/Details/StudioDetailsYear.js
@@ -271,18 +271,6 @@ class StudioDetailsYear extends Component {
 
                     {translate('Search')}
                   </MenuItem>
-
-                  <MenuItem
-                    onPress={this.onInteractiveSearchPress}
-                    isDisabled={!totalMovieCount}
-                  >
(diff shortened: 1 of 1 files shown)


sampulsar:
more code to remove,

isInteractiveSearchModalOpen
onInteractiveSearchPress
onInteractiveSearchModalClose

plz12345:
If we're completely removing, I"ll amend my PR for the Performer and Studio detail view headers.  I'd only commented them out.

Tom-Estes (approved):

PR #430, opened 2025-03-23
FIX: Remove Interactive Search from PerformerDetailsStudio

### LINKED ISSUE #428: Fix Interactive Search within Performer Details
Interactive search within a performer gets some interesting results. When doing a search for Abby Rose with a Brazzers Exxtra Scene I get GotFilled results with Blake Blossom

Gykes:
Currently Interactive search within a performer site just does nothing.

I am removing the front end as it won't be that useful anyway. Radarr doesn't have this functionality as it expects 1 search per studio (Movie). Sonarr has this ability to search per year but it searches for packs/siterips.

We could potentially make it work with siterips but would require significant work.

There is already a PR open to remove the interactive search from the header 
https://github.com/Whisparr/Whisparr/pull/419

* Fixes #428

### DIFF
--- a/frontend/src/Performer/Details/PerformerDetailsStudio.js
+++ b/frontend/src/Performer/Details/PerformerDetailsStudio.js
@@ -81,7 +81,6 @@ class PerformerDetailsStudio extends Component {
       isOrganizeModalOpen: false,
       isManageMoviesOpen: false,
       isHistoryModalOpen: false,
-      isInteractiveSearchModalOpen: false,
       lastToggledMovie: null
     };
   }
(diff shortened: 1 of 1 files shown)


sampulsar:
There should be some more code to remove, 
isInteractiveSearchModalOpen
onInteractiveSearchPress
onInteractiveSearchModalClose

Gykes:
> There should be some more code to remove, isInteractiveSearchModalOpen onInteractiveSearchPress onInteractiveSearchModalClose

The interactive search within the Year + still still uses the modal. 

Does that not use the same pages?

plz12345:
Just remove any references on the 2 files you're editing, not the modal files themselves.  You can remove the functions as well as the imports, on your edited files.

Tom-Estes (approved):

PR #434, opened 2025-03-25
New: Import scenes without adding the scene first by parsing the filename and searching skyhook

Tom-Estes:
#### Database Migration
YES - Added SceneImportFolderFormat to NamingConfig table

#### Description
Import scenes without adding the scene first by parsing the filename and searching skyhook for the matching scene. The files can be in Scene folder format: _**Scene Title - Release Date**_, or just dumped in the import folder (useful if you're coming from v2) as long as it follows the _**Studio - Date - [Performers] - Title**_ format. I tested it on naming typical of scenes from an indexer. Quality **doesn't** need to be in the title.

1. Folders needed: **_movies, scenes, import_** (you can change the name in settings > media management).
**_movies_**/{Movie Title} ({Release Year})
**_scenes_**/{Studio Title}/{Scene Title} - {Release Date}
**_import_**/
** I found it easier to weed out the files with no matches (not in StashDB) by having an import folder separate from the library. Turn on _**Delete empty folders**_ if you want to clean up as it imports (if you manually import, you may have to delete the folder after). They don't have to be named  "**_movies_**", "**_scenes_**",  and "**_import_**", but they have to be there.
2. It doesn't kick off when you reach the page, you need to click _**Add Scenes**_
3. If it doesn't find a match, it will list it as an unmapped file. You have the opportunity to add the scene manually, then go back to **_Unmapped Files_** and import.
4. Click _**Clean Unmapped**_ to clear the list but not delete the files.
5. Batch delete by selecting the files you want to delete and clicking _**Delete Selected**_.

**_Disclaimer_**: Test in batches. I've imported over 500 files at one time. The more files, the longer it takes to "spool up" before you see it start. Each time a Scene is added, it will add studios and performers also.

This should get the import ball rolling.

**TODO:** 
- Add an edit button in order to pick profiles when adding (in case you want to keep looking for an upgrade). 
- Add the rejection info as to why it returned as unmapped

image
image

#### Todos
- [x] Tests


src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -149,8 +149,14 @@ private List<UnmappedFolder> GetUnmappedFolders(string path, Dictionary<int, str
                 return results;
             }
 
-            var subFolderDepth = _namingConfigService.GetConfig().MovieFolderFormat.Count(f => f == Path.DirectorySeparatorChar);
-            var possibleMovieFolders = _diskProvider.GetDirectories(path).ToList();
+            var namingConfig = _namingConfigService.GetConfig();
+            var moviesFolderFormat = namingConfig.MovieFolderFormat;
+            var moviesFolderName = moviesFolderFormat.Split(new char[] { '\\', '/' }, StringSplitOptions.RemoveEmptyEntries)[0];
+            var moviesPath = Path.Combine(path, moviesFolderName);
```
sampulsar: Need to add the following test,as the MovieFolderFormat may not contain a \ or /. 

if (moviesFolderName == moviesFolderFormat) 
{
    moviesFolderName = string.Empty;
}

the first segment within the format could also be a token value, so then would also need to be cleaned.

sampulsar (commented):
Would need more testing with a various configurations.

{Movie Title} ({Release Year})
{Studio Title}/{Scene Title} - {Release Date}
import/

src/NzbDrone.Core/RootFolders/RootFolderService.cs
Tom-Estes: Added folder validation

Tom-Estes:
Making a new PR. Tests pass locally.

PR #438, opened 2025-03-28
Fix: studio performer performance

sampulsar:
#### Database Migration
NO

#### Description
Alter studio and performer details
- Add a delay so that the top section is able to render to the user
- Change the way that the data is filtered and sorted (createClientSideCollectionSelector sorted the complete list of scenes causing a delay)

The sorting was a major contributor to the amount of time that that it took to render these routes.

Movie Action 
- FETCH_MOVIES may be called too much, as once populated any changes for a single scene would be sent through SET_MOVIE_VALUE

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Components/Delayed.js
+++ b/frontend/src/Components/Delayed.js
@@ -1,0 +1,28 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import LoadingIndicator from 'Components/Loading/LoadingIndicator';
+
+class Delayed extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = { hidden: true };
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ hidden: false });
+    }, this.props.waitBeforeShow);
+  }
+
+  render() {
+    return this.state.hidden ? <LoadingIndicator /> : this.props.children;
+  }
+}
+
+Delayed.propTypes = {
+  waitBeforeShow: PropTypes.number.isRequired,
+  children: PropTypes.element
+};
+
+export default Delayed;
--- a/frontend/src/Performer/Details/PerformerDetails.js
+++ b/frontend/src/Performer/Details/PerformerDetails.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Alert from 'Components/Alert';
+import Delayed from 'Components/Delayed';
 import FieldSet from 'Components/FieldSet';
 import Icon from 'Components/Icon';
 import Label from 'Components/Label';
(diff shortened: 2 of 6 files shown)


plz12345:
@sampulsar I'll check out your branch and run against a large studio like Sweet Sinner.  Any particular behavior I should look out for?

sampulsar:
> @sampulsar I'll check out your branch and run against a large studio like Sweet Sinner. Any particular behaviour I should look out for?

The performance gain in mainly on the total number of scenes, and also the number of years on a studio, over the number of scenes. There should not be a huge delay between loading the studio details. There should be close to no delay rendering the top sections before the years are rendered.

You should also be able to use the arrows on studio details to navigate between studios, and a small or large studio will appear to load in the same time and no lag.

The same is for performers.

plz12345:
> > @sampulsar I'll check out your branch and run against a large studio like Sweet Sinner. Any particular behaviour I should look out for?
> 
> The performance gain in mainly on the total number of scenes, and also the number of years on a studio, over the number of scenes. There should not be a huge delay between loading the studio details. There should be close to no delay rendering the top sections before the years are rendered.
> 
> You should also be able to use the arrows on studio details to navigate between studios, and a small or large studio will appear to load in the same time and no lag.
> 
> The same is for performers.

Ack, building source for your branch now.  My test case vs. a small studio should be good.  Sweet Sinner has something like 1200 scenes since 2010 I think.

plz12345:
From my testing, it's not doing an API call when loading a studio, and I can't see any performance issues when using the arrows to switch between studios.  If I read the code correct, the delay you inserted is only 50 milliseconds, so us humans wouldn't detect that anyways.

plz12345 (approved):
From my testing, it's not doing an API call when loading a studio, and I can't see any performance issues when using the arrows to switch between studios. If I read the code correct, the delay you inserted is only 50 milliseconds, so us humans wouldn't detect that anyways.  Tested on desktop and tablet, with studios ranging from a couple scenes to 1200 over 15 years.

sampulsar:
> From my testing, it's not doing an API call when loading a studio, and I can't see any performance issues when using the arrows to switch between studios. If I read the code correct, the delay you inserted is only 50 milliseconds, so us humans wouldn't detect that anyways. Tested on desktop and tablet, with studios ranging from a couple scenes to 1200 over 15 years.

There is a delay there, so that the first components are able to render and be viewed, without the thread being rendering the scenes section, that take long enough to cause the page to have a delay.

We don't need to alter the API as when the scenes are rendered now it is filtered for that studio/year and then sorted. The sorting has a higher performance hit based on the number  of items sorted over the filtering. Changing the order then makes that section faster.

Tom-Estes (approved):

PR #444, opened 2025-03-31
New: 5K_6K_8K_Qualities Migration

Gykes:
Setting this as a draft. Is someone able to verify me please. Not quite sure how to test this without an old DB

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/006_5K_6K_8K_Qualities.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/006_5K_6K_8K_Qualities.cs
@@ -61,6 +61,19 @@ namespace NzbDrone.Core.Datastore.Migration
 
         public void UpdateQualityToQualityDefinition()
         {
+            using (var insertDefinitions = _connection.CreateCommand())
+            {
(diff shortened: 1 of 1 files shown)


plz12345:
Looks like DB checks failed, so probably need some tests updated.  This is what I slapped together on how you could test this migration and the bugfix it's intended to cover.

1. shut down local dev instance
2. rename the DB
3. `git checkout b4248696f5936b3745c3b4490dcc7d0ffe9633c5` this is the commit just before your previous change
4. build and fire it up, add some profiles
5. shut down local instance
6. `git checkout 07ce836580650c2cc416a12bf017edb644c5ed9a` this is your iinitial commit that broke editing of existing profiles.  build, start, and confirm that's the case
7. shut down local instance again, and back up that database copy to save re-testing of 1-6
8. `git checkout your-fix-branch`
9. build and fire it up, and confirm your migration works.  It should let you edit your original profiles, and include your new resolutions.  If not, shut it down, restore DB form step 7, and re-assess your fixes.

Tom-Estes:
https://github.com/Whisparr/Whisparr/pull/445

PR #446, opened 2025-04-02
New: Adding Extra Performer Naming Tokens

Gykes:
Keeping this as a draft as this seems like a lot tbh.

Opinions?


sampulsar:
@Gykes 

Don't alter the genders as that matches the data from stashDB. There is no other on StashDB.

The following will give you other (Not female, Not Male)

var otherPerformers = episodes
    .SelectMany(e => e.Actors.Where(a => a.Gender != Gender.Female && a.Gender != Gender.Male && ).Select(a => a.Name)).Distinct()
    .Take(4).ToList();
return string.Join(" ", otherPerformers);

Gykes:
So, based on our convo yesterday I didn't bother updating. 

Mostly I just want to know if we are keeping it as other or if we want tokens for each individual Gender than we have available.

sampulsar:
Not required unless requested

PR #448, opened 2025-04-03
New: Import scenes without adding the scene first by parsing the filename and searching skyhook

Tom-Estes:
#### Database Migration
YES - Added SceneImportFolderFormat to NamingConfig table

#### Description
Import scenes without adding the scene first by parsing the filename and searching skyhook for the matching scene. The files can be in Scene folder format: _**Scene Title - Release Date**_, or just dumped in the import folder (useful if you're coming from v2) as long as it follows the _**Studio - Date - [Performers] - Title**_ format. I tested it on naming typical of scenes from an indexer. Quality **doesn't** need to be in the title.

1. Folders needed: **_movies, scenes, import_** (you can change the name in settings > media management).
**_movies_**/{Movie Title} ({Release Year})
**_scenes_**/{Studio Title}/{Scene Title} - {Release Date}
**_import_**/
** I found it easier to weed out the files with no matches (not in StashDB) by having an import folder separate from the library. Turn on _**Delete empty folders**_ if you want to clean up as it imports (if you manually import, you may have to delete the folder after). They don't have to be named  "**_movies_**", "**_scenes_**",  and "**_import_**", but they have to be there.
2. It doesn't kick off when you reach the page, you need to click _**Add Scenes**_
3. If it doesn't find a match, it will list it as an unmapped file. You have the opportunity to add the scene manually, then go back to **_Unmapped Files_** and import.
4. Click _**Clean Unmapped**_ to clear the list but not delete the files.
5. Batch delete by selecting the files you want to delete and clicking _**Delete Selected**_.

**_Disclaimer_**: Test in batches. I've imported over 500 files at one time. The more files, the longer it takes to "spool up" before you see it start. Each time a Scene is added, it will add studios and performers also.

This should get the import ball rolling.

**TODO:** 
- Add an edit button in order to pick profiles when adding (in case you want to keep looking for an upgrade). 
- Add the rejection info as to why it returned as unmapped

image
image

#### Todos
- [x] Tests


src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -149,7 +149,11 @@ private List<UnmappedFolder> GetUnmappedFolders(string path, Dictionary<int, str
                 return results;
             }
 
-            var subFolderDepth = _namingConfigService.GetConfig().MovieFolderFormat.Count(f => f == Path.DirectorySeparatorChar);
+            var config = _namingConfigService.GetConfig().MovieFolderFormat;
```
sampulsar, line 152: Should this always be MovieFolderFormat or should it be conditional if the user is looking at scenes and then uses SceneFolderFormat?

src/NzbDrone.Core/Organizer/FileNameValidation.cs
```diff
@@ -29,7 +37,23 @@ public static IRuleBuilderOptions<T, string> ValidSceneFolderFormat<T>(this IRul
...
-            return ruleBuilder.SetValidator(new RegularExpressionValidator(FileNameBuilder.SceneTitleRegex)).WithMessage("Must contain scene title");
+            return ruleBuilder.SetValidator(new RegularExpressionValidator(FileNameBuilder.SceneFolderRegex)).WithMessage("Must contain scene title");
+        }
+
+        public static IRuleBuilderOptions<T, string> ValidMainSceneFolderFormat<T>(this IRuleBuilder<T, string> ruleBuilder)
+        {
+            ruleBuilder.SetValidator(new NotEmptyValidator(null));
+            ruleBuilder.SetValidator(new IllegalCharactersValidator());
+
+            return ruleBuilder.SetValidator(new RegularExpressionValidator(FileNameBuilder.MainSceneFolderRegex)).WithMessage("Must contain scenes folder");
```
sampulsar: Do you need these validations as it can cause people to change their Folders?

The following does not validate, 
{Studio Network}/{Studio Title}/{Release-Date} - {Scene Title} [{StashId}]

Requires scenes at the front, and {Studio Network} needs to be removed.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -189,9 +193,24 @@ private List<UnmappedFolder> GetUnmappedFolders(string path, Dictionary<int, str
             return results.OrderBy(u => u.Name, StringComparer.InvariantCultureIgnoreCase).ToList();
```
sampulsar, line 193: Do you want to limit the amount of results it finds, so it is not hitting the disk for 10,000+ directories

src/NzbDrone.Core/Organizer/FileNameValidation.cs
Tom-Estes: > {Studio Network}/{Studio Title}/{Release-Date} - {Scene Title} [{StashId}]

Would need to be scenes/{Studio Network}/{Studio Title}/{Release-Date} - {Scene Title} [{StashId}]

Main reason was to only add one root folder. No matter if you're adding a scene or movie, you wouldn't need to change the root.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
Tom-Estes: Tried both ways. It would be better to determine which format to use earlier in the process, but that would be a different PR.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
Tom-Estes: Honestly never looked at it. Been running it production for about 8 months and haven't had an issue.

src/NzbDrone.Core/Organizer/FileNameValidation.cs
Tom-Estes: I changed the regex to only look for folder/. Since it is looking for the main folder only, it should pass the folder validation.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
sampulsar: Maybe a lock on the function, so it does not run too much, another PR can clean this up.

sampulsar:
Do you want to retest this after correcting the unit tests as it looks to be one folder down, as the first segment is stripped, and not used?

sampulsar (approved):
Could not fully test, but did not break the environment

PR #462, opened 2025-04-11
Fix: local scene covers

sampulsar:
#### Database Migration
NO

#### Description
Update scene covers with the local URL

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/src/Whisparr.Api.V3/Movies/MovieController.cs
+++ b/src/Whisparr.Api.V3/Movies/MovieController.cs
@@ -192,6 +192,7 @@ namespace Whisparr.Api.V3.Movies
             var moviesResources = new List<MovieResource>();
 
             var movieStats = _movieStatisticsService.MovieStatistics(ids);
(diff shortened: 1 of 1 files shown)


plz12345:
@sampulsar , I tested this locally and it works great.  I'm sure the StashDB admin team will appreciate the load reduction on their server/CDN.  It does try load the full size images, though.  

Since the app doesn't really need the original StashDB size in such small layouts in Whisparr, I suggest adding the following edits to this PR.  This will load the smaller versions of these covers.  Detail view would load the 360px one and overview/poster views will use the 180px version.  I didn't dig in to figure out why the `size` param ends up being multiplied by 2 when attempting to load the image by name.

`MovieDetails.js` line 328
Change `250` to `180`

`SceneIndexOverview.tsx` line 153
Change `250` to `90`

`SceneIndexPoster.tsx` line 182
Change `250` to `90`

Tom-Estes (approved):

PR #465, opened 2025-04-13
Add: Validate Runtime Parameter

sampulsar:
#### Database Migration
NO

#### Description
Allow the runtime validation to be turned on or off. 
Corrected when the minutes are returned as 0 on some videos

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/.vscode/settings.json
+++ b/frontend/.vscode/settings.json
@@ -9,7 +9,7 @@
 
   "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
-    "source.fixAll": true
+    "source.fixAll": "explicit"
   },
 
   "typescript.preferences.quoteStyle": "single",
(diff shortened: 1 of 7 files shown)


stayrusty:
In this PR, you might want to fix this, as it's related to the process.

https://github.com/Whisparr/Whisparr/blob/926aecd3660451bfe50d101ded59ba7d44af0b30/src/NzbDrone.Core/MediaFiles/MovieImport/ImportDecisionMaker.cs#L273-L279

Where you're using `localMovie.MediaInfo.RunTime.Minutes`, you need to use `localMovie.MediaInfo.RunTime.TotalMinutes` and then round it. `Minutes` only gets the minutes component (0-59). Resulting in: 

image

If the runtime is over an hour.

PR #466, opened 2025-04-13
Development Override MetadataURL for testing newer schema

sampulsar:
#### Database Migration
NO

#### Description
To test development of skyhook schema changes, etc, and then able to test with the released version. 

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
@@ -61,6 +61,7 @@ namespace NzbDrone.Core.Configuration
         string PostgresMainDb { get; }
         string PostgresLogDb { get; }
         string Theme { get; }
+        string WhisparrMetadata { get; }
     }
 
     public class ConfigFileProvider : IConfigFileProvider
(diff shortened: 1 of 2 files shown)


plz12345:
@sampulsar it looks like you are using a configuration file, vs. DB config entry, for this one.  How can we review/test it?  I think we'd need an alternate endpoint, or at least we'd need to know where the new config file lives and schema for it (to see the attempts to use the new endpoint in logs).

sampulsar:
> @sampulsar it looks like you are using a configuration file, vs. DB config entry, for this one. How can we review/test it? I think we'd need an alternate endpoint, or at least we'd need to know where the new config file lives and schema for it (to see the attempts to use the new endpoint in logs).

C:\ProgramData\Whisparr\config.xml is the standard location and the new item is automatically added to the file

<WhisparrMetadata>https://api.whisparr.com/v4/{route}</WhisparrMetadata>

plz12345:
Gotcha.  Noted, looks good to me.

plz12345 (approved):

PR #473, opened 2025-05-13
Fix: Remove cache item

sampulsar:
#### Database Migration
NO

#### Description
Find the correct MovieResource cache so that the items are removed then updated or deleted

#### Todos
- [x] Tests

* Fixes #XXXX


plz12345 (approved):
Tested locally.  Delete does delete, in DB and now also the cache.

PR #476, opened 2025-05-14
Fix: Studio Aliases

sampulsar:
#### Database Migration
NO

#### Description
Fix Aliases and AfterDate updating

#### Todos
- [x] Tests

* Fixes #XXXX


sampulsar:
Test by adding a scene for a new studio
Test by altering that studio, unmonitor

plz12345:
@sampulsar what behavior are we expecting to see?  I'm not really sure I even see anything different.

1. Added `b1b5df02-8c61-4946-b9fc-ae70c795a838`
2. Went to studio and added a tag, saved
3. Un-monitored scene


That is does not error adding the new studio, and that the studio can be modified without the after date set.

plz12345 (approved):
Unable to repro any NPE's when searching scenes, so I think this is good.

PR #522, opened 2025-06-16
New: Adding Scene Code as Naming Token

### LINKED ISSUE #472: Add "ID" from TPDB and "Studio Code" from StashDB as options during custom renames.

Currently, the best way I've found to include the studio code is to manually include it in the importing file and use it as the release group tag, but this causes issues when the file also has a release group. 

I would like to be able to add the studio ID/code (not the Stash ID) to the filename when Whisparr automatically renames the files. It just needs to be added to the template as an option. 

The other alterternative I considered was using the StashID, which is much too long to be useful when appending to a filename.

### Anything else?

_No response_

Gykes:
Adds in the ability to use the scene code as a naming token. If the token is Null from Skyhook then it will just be blank.

* Fixes #472

### DIFF
--- a/frontend/src/Settings/MediaManagement/Naming/NamingModal.js
+++ b/frontend/src/Settings/MediaManagement/Naming/NamingModal.js
@@ -92,7 +92,8 @@ const sceneTokens = [
   { token: '{Scene PerformersFemaleAlias}', example: 'Female Performers (Alias)' },
   { token: '{Scene PerformersMaleAlias}', example: 'Male Performers (Alias)' },
   { token: '{Release Date}', example: '2009-02-04' },
-  { token: '{Release ShortDate}', example: '09 02 04' }
+  { token: '{Release ShortDate}', example: '09 02 04' },
+  { token: '{Scene ID}', example: '12345' }
 ];
 
 const studioTokens = [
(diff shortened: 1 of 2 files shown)


frontend/src/Settings/MediaManagement/Naming/NamingModal.js
```diff
@@ -92,7 +92,8 @@ const sceneTokens = [
   { token: '{Scene PerformersFemaleAlias}', example: 'Female Performers (Alias)' },
   { token: '{Scene PerformersMaleAlias}', example: 'Male Performers (Alias)' },
   { token: '{Release Date}', example: '2009-02-04' },
-  { token: '{Release ShortDate}', example: '09 02 04' }
+  { token: '{Release ShortDate}', example: '09 02 04' },
+  { token: '{Scene ID}', example: '12345' }
```
plz12345: I feel like this should be `Studio Code` not `Scene ID`.  It'd hold true to what Stash calls it, and to the DB column naming.  `Scene ID` may be confused for StashDB ID (GUID).

image

Gykes: The reason I am making it Scene ID is because the verbiage is just extremely confusing. StashDB uses Studio Code as the code that a studio gives the scene (Scene ID) and it is not the code for the studio. That code that you highlighted is actually the Scene ID/Scene Code for the scene. Naming it Studio Code is just going to cause confusion. 

Q: "I have Studio Code in my naming, why are all the codes different" 
A:"Well the Studio Code isn't actually the studio code... it's the scene code"

sampulsar (approved):

PR #525, opened 2025-06-19
New: Add tags to exclusion list

plz12345:
#### Database Migration
NO

#### Description
Users may now manually add `Tags` to `Settings > Import Lists > Import List Exclusions`.  At present, the backend code is looking at case-insensitive tag names due to limitations in the v4 Skyhook API.  Expectation is that if  Skyhook can be updated, the backend code doing the comparisons can be easily shifted to compare by foreignId without any changes needed by the user.

I also slightly tweaked the log formatting to be more legible for all exclusion types.

#### Todos
- [X] Tests

### DIFF
--- a/src/NzbDrone.Core/Movies/AddMovieService.cs
+++ b/src/NzbDrone.Core/Movies/AddMovieService.cs
@@ -227,7 +227,7 @@ namespace NzbDrone.Core.Movies
                     if (excludedStudio)
                     {
                         _importExclusionService.AddExclusion(newExclusion);
-                        throw new ValidationException($"Studio: {newMovie.MovieMetadata.Value.Studio.Title} has been excluded");
+                        throw new ValidationException($"Studio: [{newMovie.MovieMetadata.Value.Studio.Title}] has been excluded");
                     }
                     else
                     {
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Movies/AddMovieService.cs
```diff
@@ -252,9 +252,18 @@ private Movie SetPropertiesAndValidate(Movie newMovie)
                     if (excludedPerformers.Any())
                     {
                         _importExclusionService.AddExclusion(newExclusion);
-                        throw new ValidationException($"Performer: {string.Join(",", excludedPerformers.Select(ep => ep.MovieTitle).ToList())} has been excluded");
+                        throw new ValidationException($"Performer: [{string.Join(",", excludedPerformers.Select(ep => ep.MovieTitle).ToList())}] has been excluded");
                     }
                 }
+
+                var tagNames = newMovie.MovieMetadata.Value.Genres.ToList();
+                var excludedTags = _importExclusionService.GetAllByType(ImportExclusionType.Tag).Where(e => tagNames.Contains(e.MovieTitle, StringComparer.OrdinalIgnoreCase)).ToList();
```
sampulsar: _importExclusionService.GetAllByType(ImportExclusionType.Tag) can be null and causes an exception within the unit tests

sampulsar (commented):
Fix null exception

src/NzbDrone.Core/Movies/AddMovieService.cs
plz12345: > _importExclusionService.GetAllByType(ImportExclusionType.Tag) can be null and causes an exception within the unit tests

That function's return could be null in general, not just for Tag, so I fixed it in the original class by always returning a List instead of allowing it to return null.  Hope that's OK.

sampulsar (approved):

PR #526, opened 2025-06-21
Fix: Tag exclusion output logging

plz12345:
#### Database Migration
NO

#### Description
My previous PR had a small bug in the output logging, so it was logging the names of the configured tag exclusions, not the ones that matched the scene's genres array.

#### Todos
- [x] Tests


sampulsar (approved):

sampulsar (commented):
Change "Pulp Fiction" to "The Dark Knight", as Radarr is using that for the test and it will pass. should_be_able_to_update_multiple_movies

plz12345:
@sampulsar isn't Whisparr limiting TMDB queries to the adult genre?  I switched it to "Sinner", which does pass.

PR #535, opened 2025-07-09
New: Adding Release-ShortDate Naming Token

Gykes:
#### Description
Adding in `Release-ShortDate` 

eg: `25-01-01` instead of `2025-01-01`

EDIT: 
I removed `Release Date` as it was not parsing out the - anyway and was broken. 
https://github.com/Whisparr/Whisparr/issues/536

Created an issue for it here and will look into fixing it in a separate PR


Gykes:
Nevermind, decided not to be lazy and just fix it first.

PR #549, opened 2025-07-24
Chore: MetadataURL changes

Gykes:
Will allow someone that is selfhosting to point their client at their local version via a config setting in `Config.xml`


plz12345 (approved):
These look good.  Only thing I could think of to potentially add is coverage for strings that are not URL's, but normies aren't tinkering with this in the first place.

Gykes:
Fair enough, could potentially do a fallback to default API if the string is not a valid URL. If it becomes a problem in the future we can adjust

PR #555, opened 2025-08-01
New: Indexer Search Changes

Gykes:
Firstly, I wanted to note I am kind of against this. People can abuse it and get bad matches/banned from their trackers if they just spam API requests all day long. If you have all options selected it can essentially send 7 requests for 1 scene. Users should know that they should only really select what is needed to help prevent tracker bans

Adds the ability to change the individual indexer searches. I made it individual as some trackers have better naming standards than others and no point making unnecessary API calls if the naming is "normal". This, for the most part, is the same setup as V3 currently has.

- Site + Title only
- Title Only
- yy.mm.dd, dd.mm.yy or both date variatioinis
- Use the Site, Network, or both

This is tucked away behind the advanced features inside of Newznab and Torznab

Edit: Users should be aware that just because a search was found and downloaded does not mean that it can be imported. A lot of things will need to be manually imported due to not having the correct formatting.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


plz12345 (approved):
This looks solid, and good use of additional unit testing.

PR #567, opened 2025-08-05
Fix: Sites With > 1 Release a Day

Gykes:
The search functionality was breaking when sites released multiple scenes per day. The root issue was that `EpisodeRepository.FindOneByAirDate()` would crash with an `InvalidOperationException` when multiple episodes shared the same air date, and the episode matching logic wasn't conservative enough to prevent incorrect downloads. Additionally, interactive searches were bypassing validation entirely because they only contained a single episode in the search criteria, allowing incorrect releases to be accepted even when multiple episodes existed for that date in the database.

I implemented conservative episode matching that requires definitive scene title matches for multi-episode dates. Key changes
include:

- Fixed the crash in `EpisodeRepository` to return the first episode with logging instead of throwing exceptions 
- Enhanced `EpisodeService.FindOneByAirDate()` with matching logic that compares scene titles, performer names, and handles punctuation variations while rejecting ambiguous matches.
- Fixed search timing conflicts by adding unique millisecond offsets to `baseSearchTime` for multiple episodes
- Most importantly, updated `ParsingService.GetDailyEpisode()` to always validate releases against the database using conservative matching logic, even when search criteria contains a specific episode.

The system now prevents incorrect downloads by rejecting releases that don't definitively match a specific episode's scene title when multiple episodes exist for the same date, while maintaining full functionality for single-episode-per-day series. Both automatic and interactive searches now apply the same conservative validation, ensuring that releases like `BrazzersExxtra.25.07.18.Lexi.Lore.Getting.Into.Double.Trouble` are properly rejected when searching for `Brazzers Exxtra-2025-07-18-Slutty Sci-Fi Celeb-Litty Britty, Juan El Caballo Loca` instead of being incorrectly accepted.

Personal statement: I'm just a dude with too much time on his hands, if there are better ways to handle this then please educate/point me in the correct direction and I can take a look.

### DIFF
--- a/src/NzbDrone.Core.Test/TvTests/EpisodeServiceTests/ByAirDateFixture.cs
+++ b/src/NzbDrone.Core.Test/TvTests/EpisodeServiceTests/ByAirDateFixture.cs
@@ -91,7 +91,7 @@ namespace NzbDrone.Core.Test.TvTests.EpisodeServiceTests
         }
 
         [Test]
-        public void should_get_episode_when_two_regular_episodes_with_perfomer_in_title_share_the_same_air_date_and_performer_and_part_is_provided()
+        public void should_return_null_when_two_regular_episodes_with_perfomer_in_title_share_the_same_air_date_and_performer_causes_ambiguous_match()
         {
             var title1 = "EvilAngel - 2021-02-24 - Whitney Wright POV Anal & A2M";
             var title2 = "EvilAngel - 2021-02-24 - BTS Whitney Wright POV Anal & A2M";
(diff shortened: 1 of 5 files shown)


Gykes:
Forgot to add in my last commit to the PR. 

That commit changes it from validating ALL scenes to only scenes that have > 1 release per day. The original logic for single release works well and this will limit unnecessary DB lookups.

Gykes:
My new logic returns a NULL value if it does not pass the >1 scene per date. Updated the tests to expect a null value.

plz12345 (approved):
Tested locally, and works exactly as desired.

PR #568, opened 2025-08-05
Fix: Environment Variables

### LINKED ISSUE #520: SSL envs do not work

### Current Behavior

Hello, I am using the hotio/whisparr image to host this app. I have noticed that the envs documented here do not work for SSL
https://wiki.servarr.com/en/useful-tools

Image

However, the postgres envs did work. I did open the issue here because this should be an upstream issue instead of an issue with the image.

```
      - WHISPARR__SERVER__ENABLESSL=true
      # - WHISPARR__SERVER__PORT=8008 # SSL port
      - WHISPARR__SERVER__SSLCERTPATH=/certs/cert.p12
```

### Expected Behavior

it should enable the corresponding settings when it sees the envs.

### Steps To Reproduce

1. download ghcr.io/hotio/whisparr:latest
2. deploy the image with the above stated envs
3. check the ssl status on the UI, it is still disabled

### Environment

```markdown
- OS: Ubuntu 24.04
- Whisparr: 2.0.0.1112
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: not UI
- Database: Postgres
```

Develop (V2)

### Trace Logs?

No logs

### Anything else?

_No response_

Gykes:
Environment variables for server configuration (WHISPARR__SERVER__*) were being ignored, causing Docker deployments and environment-based configuration to fail for basic server settings.

Bootstrap.cs was reading server configuration using flat property names ("EnableSsl", "Port", etc.) instead of the hierarchical configuration paths ("Whisparr:Server:EnableSsl", "Whisparr:Server:Port", etc.) that Microsoft.Extensions.Configuration creates from environment variables.

I am unable to test this so will request that the original requester test this before merging. Please don't merge until proper testing is done.

Fixes: #520

### DIFF
--- a/src/NzbDrone.Host/Bootstrap.cs
+++ b/src/NzbDrone.Host/Bootstrap.cs
@@ -125,12 +125,12 @@ namespace NzbDrone.Host
         {
             var config = GetConfiguration(context);
 
-            var bindAddress = config.GetValue(nameof(ConfigFileProvider.BindAddress), "*");
-            var port = config.GetValue(nameof(ConfigFileProvider.Port), 6969);
-            var sslPort = config.GetValue(nameof(ConfigFileProvider.SslPort), 8008);
-            var enableSsl = config.GetValue(nameof(ConfigFileProvider.EnableSsl), false);
-            var sslCertPath = config.GetValue<string>(nameof(ConfigFileProvider.SslCertPath));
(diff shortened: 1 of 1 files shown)


plz12345 (approved):
This looks good, using existing config as a fallback rather than the only option.  Should be fine for Docker and bare metal installs.

PR #573, opened 2025-08-07
Fix: No rejections to the Top in Interactive Search

### LINKED ISSUE #572: V2: Interactive Search Results Should Filter Passing Results First

### Current Behavior

This is mostly for tracking purposes otherwise I will forget. 

Currently interactive search does not put passing results to the top causing users to have to scroll through the entire list to view any potential passed releases near the bottom.

### Expected Behavior

All passed releases should be filtered to the top. There is already filtering in place to filter by release so perhaps setting it to asc by default is the best solution 

### Steps To Reproduce

Use interactive search.

### Environment

```markdown
sadf
```

Develop (V2)

### Trace Logs?

sadf

### Anything else?

_No response_

Gykes:
Currently, rejected releases get mixed in with approved ones in the interactive search. This will make it so that releases with 0 rejections get put above release with greater than 0 rejections.

Fixes: #572

### DIFF
--- a/frontend/src/Store/Actions/releaseActions.js
+++ b/frontend/src/Store/Actions/releaseActions.js
@@ -57,6 +57,19 @@ export const defaultState = {
       }
 
       return releaseWeight;
+    },
+
+    releaseWeight: function(item, direction) {
+      const rejections = item.rejections;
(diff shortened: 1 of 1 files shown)


plz12345 (approved):
Nice work

PR #586, opened 2025-08-11
Fix: Existing Only Monitor

### LINKED ISSUE #64: V2: Monitor Existing Scenes still pulls new scenes

### Current Behavior

Under site monitoring, monitor existing, i would expect, just to monitor the existing files not new releases, but its pulling all future releases too.

### Expected Behavior

Just monitor existing releases, not future releases

### Steps To Reproduce

1. Go to any site
2. click site monitoring
3. select existing scenes

### Environment

```markdown
- OS: Unraid 6.11.5
- Whisparr Version: 2.0.0.146
- Package Version: nightly-321d657 by hotio 
- .Net Version: Yes (6.0.13)
- Database: Sqlite 3.38.5
- Docker Install: yes
- Using Reverse Proxy:no
```

Main

### Trace Logs?

[whisparr.trace.txt](https://github.com/Whisparr/Whisparr/files/11168313/whisparr.trace.txt)

### Anything else?

If the correct behavior of 'monitor existing' is to both monitor existing and future. A Monitor existing only option would be immensly helpful

Gykes:
When "Monitor Existing" was selected, it would only monitor episodes with files initially. However, when new episodes were added to the system later (via episode refresh), they were automatically monitored regardless of whether they had files, because the logic only checked `MonitorNewItems` and season monitoring status.

@plz12345 This is kind of a YOLO I don't have a good way to test this. I don't really feel like waiting for a week or so for a release to hit and test. 

* Fixes #64

### DIFF
--- a/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
+++ b/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
@@ -112,6 +112,12 @@ namespace NzbDrone.Core.Tv
                 return false;
             }
 
+            // If the series was set to monitor "Existing" only, don't monitor new episodes without files
+            if (series.AddOptions != null && series.AddOptions.Monitor == MonitorTypes.Existing && !episode.HasFile)
+            {
+                return false;
+            }
+
             var season = seasons.SingleOrDefault(c => c.SeasonNumber == episode.SeasonNumber);
             return season == null || season.Monitored;
         }
--- a/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
+++ b/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
@@ -113,7 +113,7 @@ namespace NzbDrone.Core.Tv
             }
 
             // If the series was set to monitor "Existing" only, don't monitor new episodes without files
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
```diff
@@ -112,6 +112,12 @@ private bool GetMonitoredStatus(Episode episode, IEnumerable<Season> seasons, Se
                 return false;
             }
 
+            // If the series was set to monitor "Existing" only, don't monitor new episodes without files
+            if (series.AddOptions != null && series.AddOptions.Monitor == MonitorTypes.Existing && !episode.HasFile)
```
plz12345: @Gykes this can be simplified and made null-safe at the same time

```
if (series.AddOptions?.Monitor == MonitorTypes.Existing && !episode.HasFile)
{
    return false;
}
```

Gykes: Updated

plz12345 (approved):

sampulsar (approved):

PR #595, opened 2025-09-11
New: JAV - Search and Import Support

Gykes:
There are a lot of changes here. 

With the implementation of JAV and External ID support then this is the next logical step. 

Adds: 
- DB migration for External ID column
- DB Index for External ID to keep the External Id search as fast as possible
- External ID Testing
- Search via External ID for Torrents and Usenet
- Match via External ID for searching and importing (regexes).
- Fall back to normal Site + date matching if no external ID is found


Gykes:
Going to break this PR up. Lots of changes and additions that should be separated out for easier testing and tracking

PR #599, opened 2025-09-16
New: Add External ID Support

Gykes:
This will allow Whisparr to pull down and store the `ExternalId` of all episodes. 

Added: 
- DB migration for the new ExternalId column
- Store External ID on refresh of series.

Will be used for JAV in future PR's


plz12345 (approved):
Looks good

PR #600, opened 2025-09-16
Fix: Truncate Long Filenames

Gykes:
Whisparr only truncated release tokens individually and not the entire filename causing errors. If the total filename was > 255 but the title, or any other token, was < 255 it would attempt to rename and then error out as the OS would not allow it. 

This new feature checks the total length of the filename, not just individual tokens, and if > 255 starts to truc the title (Assuming titles will always be the longest) till the overall filename drops below 255.


plz12345:
@Gykes can you add testes for this use case, the existing test fixture?

Gykes:
> @Gykes can you add testes for this use case, the existing test fixture?

Building

plz12345 (approved):
Looking good

PR #608, opened 2025-09-20
New: Filter and sort Performers by age, career start, career end

### LINKED ISSUE #607: V3: requesting updated ui sortable fields for performers

  It would be nice to be able to sort and filter these fields for performers.
        public int? Age { get; set; }
        public int? CareerStart { get; set; }
        public int? CareerEnd { get; set; }
        public PerformerStatus Status { get; set; }

 https://github.com/Whisparr/Whisparr/blob/eros/src/NzbDrone.Core/Movies/Performers/Performer.cs

NA

### Anything else?

_No response_

plz12345:
#### Database Migration
NO

#### Description
- Added ability to filter and sort by performer age, career start, career end
- Added ability to filter by performer active/inactive status
- Added Age and Status to performer detail view

#### Notes
- Age is based on numerical value from Whisparr API, not calculated based on date of birth.  Whisparr doesn't fetch that.
- Status was added as filter, only.  It caused issues as a sortable column, and the data isn't very accurate anyways since StashDB does not always have accurate info on when a performer retires, comes out of retirement, etc.
- This looks like a lot more changes than reality, due to IDE code formatting around changed lines.

image

image

image

#### Todos
- [x] Tests

* Fixes #607


frontend/src/Performer/Details/PerformerDetails.css.d.ts
```diff
@@ -1,6 +1,7 @@
 // This file is automatically generated.
 // Please do not change this file!
 interface CssExports {
+  'age': string;
```
Gykes, line 4: Would age not be better as an int?

frontend/src/Performer/Details/PerformerDetails.js
```diff
@@ -599,6 +561,7 @@ PerformerDetails.propTypes = {
   foreignId: PropTypes.string,
   fullName: PropTypes.string.isRequired,
   gender: PropTypes.string,
+  age: PropTypes.number,
```
Gykes, line 564: Age is a `.number` here but string above, not sure if that makes a difference.

frontend/src/Performer/Index/Table/PerformerIndexRow.css.d.ts
```diff
@@ -2,13 +2,17 @@
...
+  'careerEnd': string;
+  'careerStart': string;
   'cell': string;
   'checkInput': string;
   'ethnicity': string;
   'externalLinks': string;
   'fullName': string;
   'gender': string;
   'hairColor': string;
+  'monitored': string;
```
Gykes, line 15: Would this be better as a bool? I assume this is meaning if the performer is monitored so True/False.

Gykes (approved):

PR #612, opened 2025-09-21
New: Parse and import by external

Gykes:
Update from previous PR. 

Should allow users to import based on the external ID of the release name. Still testing, will update this block after some more testing is done.


Gykes:
Closing this as I need to find a better way to handle this. I'm half tempted to remove the JAV support completely unless users actually use it.

PR #658, opened 2025-11-12
New: Cache Performers

sampulsar:
#### Database Migration
NO

#### Description
Enable the cache on the performers.
Fix saving WhisparrValidateRuntime 
Improve performance of movieController

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Settings/General/WhisparrSettings.js
+++ b/frontend/src/Settings/General/WhisparrSettings.js
@@ -17,6 +17,7 @@ function WhisparrSettings(props) {
   const {
     whisparrAutoMatchOnDate,
     whisparrCacheMovieAPI,
+    whisparrCachePerformerAPI,
     whisparrValidateRuntime
   } = settings;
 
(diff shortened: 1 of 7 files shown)


plmcgrn:
A couple questions on this caching:

1. It looks like the new Set implementations are not passing a lifetime/duration.  If a user (or the scheduled task) refreshes, is it replacing/negating the cache for these items?
2. Presuming the performance impact on this is avoiding disk reads.  Will users on SSD even see a difference?
3. For the movie controller, I see you're chunking the database reads.  Is this still blocking for the overall API call (get all movies), or will it chunk those responses to the UI, too?

sampulsar:
1. The "PerformerService" removes the resource, e.g. cache.
2. The impact is dependant on the size and number of performer resources, as there is CPU and DISK IO for each resource.
3. Chunking reduces the blocking time, as other calls that may access similar objects within the system/database can process. It does not block for the duration of the API call. There are also joins that can consume memory within the database, and a chuck limits the impact.

plz12345 (approved):

PR #686, opened 2025-11-24
Eros code

sampulsar:
#### Database Migration
NO

#### Description
Match using the Episode with Studio Code
Show more information on the Parser Test
Search by Code
Fix the StashId match

#### Todos
- [x] Tests

* Fixes #XXXX


src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
```diff
@@ -117,27 +117,29 @@ public async Task<List<DownloadDecision>> MovieSearch(Movie movie, bool userInvo
                     }
                 }
 
+                var code = sceneSearchSpec.Movie.MovieMetadata.Value.Code;
```
plz12345, line 120: @sampulsar I think this should be nullsafe, since Studio Code isn't a required field in StashDB, and a large number of scenes don't have one.

plz12345 (changes requested):

src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
sampulsar: @plz12345 
when the value is used the null of whitespace is checked, so it does not use this.

if (_configService.SearchStudioCode && code.IsNotNullOrWhiteSpace())

plz12345: right, but wouldn't it generate a NRE prior to that, if metadata was null (lazy load)

sampulsar: It would have been throwing an error earlier as studio is read out of the metadata

plz12345 (approved):

PR #703, opened 2025-11-29
Fix: Refactor StashDB Import Lists

sampulsar:
#### Database Migration
NO

#### Description
Use the Base StashDB to use the standard settings

#### Todos
- [x] Tests

* Fixes #XXXX


plz12345:
Is the goal to inherit from StashDbImportBase.cs and the other Base classes?  Just want to ensure I understand, vs. blindly commit.  Is every API call using the same constraints, like pageSize, etc?

sampulsar:
The base had the same constant values, so that will be used instead of having each of the imports declaring them.

Only the constants that are different are now declare in the imports.

plz12345 (approved):

PR #726, opened 2025-12-05
Fix: Parse 4k/5k/6k/8k resolutions

plz12345:
#### Database Migration
YES

#### Description
- Add / Fix 4k/5k/6k/8k parsing by filename
- Add 4k/5k/6k/8k parsing based on resolution (if scanning media is enabled in settings)

image

#### Todos
- [X] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Common.Test/CacheTests/CachedFixture.cs
+++ b/src/NzbDrone.Common.Test/CacheTests/CachedFixture.cs
@@ -80,7 +80,6 @@ namespace NzbDrone.Common.Test.CacheTests
         }
 
         [Test]
-        [Platform(Exclude = "MacOsX")]
         public void should_honor_ttl()
         {
             var hitCount = 0;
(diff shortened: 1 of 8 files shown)


sampulsar:
need a data migration line 006_5K_6K_8K_Qualities.cs for WEBDL3384p

plz12345:
Not sure I understand why.  I'm not changing any, I'm just adding one.  The app inserts missing quality definitions on launch via QualityDefinitionService.InsertMissingDefinitions()

My understanding on migrations was that it was for schema changes (not this) or data fix-ups (also not this).

image

And in SQLite (don't mind my titles, I changed them via the UI post-start).  Quality ID 36 is the new one, which was inserted on app start.  That method exists in case the table was totally empty or was missing one, from what I saw in the code.
image

I know the build errors are likely test failures due to poor legacy parsing, like one I saw expecting 1080p when a title had 4k.  I'll fix all those when I wake up.

sampulsar:
The migration adds the new quality to the existing profiles.  3384p would be missing

plz12345:
OK, I think I'm almost there.  I created a new migration and inherited the code from the original 006 migration.  It does do the migration and inserts the new quality to existing profiles, but they sort unexpectedly in the UI.   I'm going to check in my current commit on the new migration, so maybe you can sanity check what I'm missing?  Personally, I can live with it, but I hate letting the technology defeat me.

The sort order when creating a new profile is as expected.

image

sampulsar (approved):

sampulsar (approved):

sampulsar (approved):

PR #729, opened 2025-12-05
New: server-side caching for Studios

### LINKED ISSUE #727: [V3] Add server-side caching for Studios

Similar to performers and Movies, Studios could use this same memory cache treatment to aid in performance.

Side note, at some point we should consider enabling these settings by default.

Image

Studio records are memory cached the same way that performers and movies are.

n/a

### Anything else?

_No response_

sampulsar:
#### Database Migration
NO

#### Description
Cache the API for the studio controller

#### Todos
- [x] Tests

* Fixes #727

### DIFF
--- a/src/NzbDrone.Core/Movies/Studios/StudioService.cs
+++ b/src/NzbDrone.Core/Movies/Studios/StudioService.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using NzbDrone.Common.Cache;
 using NzbDrone.Core.Messaging.Events;
 using NzbDrone.Core.Movies.Studios.Events;
 using NzbDrone.Core.Parser;
@@ -13,6 +15,7 @@ namespace NzbDrone.Core.Movies.Studios
         List<Studio> GetStudios(IEnumerable<int> studioIds);
         Studio GetById(int id);
         Studio FindByForeignId(string foreignId);
+        List<Studio> FindByForeignIds(List<string> foreignIds);
         List<Studio> SearchStudios(string query);
         List<Studio> GetAllStudios();
         List<string> AllStudioForeignIds();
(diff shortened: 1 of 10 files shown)


src/NzbDrone.Core/Movies/Studios/StudioService.cs
```diff
@@ -114,5 +136,14 @@ public List<string> AllStudioForeignIds()
         {
             return _studioRepo.AllStudioForeignIds();
         }
+
+        private void RemoveStudioResourcesCache(string foreignId)
+        {
+            var movieResourcesCache = _cacheManager.FindCache(_cacheName);
```
plz12345: should be `studio`, not `movie`, right?

src/Whisparr.Api.V3/Studios/StudioController.cs
```diff
@@ -170,5 +192,71 @@ private void LinkMovies(StudioResource resource, List<Movie> movies)
...
+                }
+            }
+
+            if (getIds.Count > 0)
+            {
+                try
+                {
+                    _studioResourceCache.Lock.Wait();
+
+                    if (getIds.Count > 0)
```
plz12345, line 247: I don't think this is needed since it's the same condition to enter the branch.

sampulsar: fixed

plz12345: I meant the repeated if statement `if (getIds.Count > 0)`.  Unless this is already removed and GH is showing me stale lines.

sampulsar:
Should be fixed not the check after the lock was there so that it can check if the previous process has already processed that recourse. 

Was not implemented for the other controllers so it is not there for all.

plz12345 (approved):

PR #748, opened 2025-12-09
Fix: cache Movie Exclusions for stasharr

sampulsar:
#### Database Migration
NO

#### Description
Cache the exclusion list for stasharr as the Development is MIA.

Stasharr is the only place using this legacy endpoint. 

Revet a change in Disk Service so that it only reads the files metadata if the file changes, but allow manual to force a redetection.

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Settings/General/WhisparrSettings.js
+++ b/frontend/src/Settings/General/WhisparrSettings.js
@@ -16,6 +16,7 @@ function WhisparrSettings(props) {
 
   const {
     whisparrAutoMatchOnDate,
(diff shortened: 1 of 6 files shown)


plz12345:
@sampulsar what was the change to disk scanning for, in relation to import list caching?

sampulsar:
Revert a change in Disk Service so that it only reads the files metadata if the file size changes, but allow manual to force a redetection.

sampulsar:
removed from this PR

plz12345 (approved):

PR #757, opened 2025-12-11
New: Parse and Import JAV

Gykes:
I finally sat down and got JAV to parse and import. 

Will do some better testing one it builds.


Gykes:
Once builds are fixed for mac I will kick this one off again

Gykes:
@sampulsar @plz12345 Was able to test and get various JAV to be found via search, download, then properly import automatically. As long as builds pass it should be g2g

sampulsar (approved):

PR #765, opened 2025-12-12
New: Update Performer Import to Select Sites

Gykes:
Currently, there is a large pain point with performer imports. You either import ALL sites that the performer belongs to or you just don't get any. This PR adds a new simple modal to allow users to control what sites they get added to their instance.

The first picture is the new button on the existing modal. I chose the verbage `Preview Sites` but it can be changed to something like `Select Sites` or something like that if people prefer it. When you click this button it opens the new modal and allows users to select and deselect what sites they want added.

The second picture is the new modal. It will list all the sites that said performer is part of. You can go through and select which one you want and it will ONLY import those sites. It will add all the other sites to a hidden exclusion list so that the import list doesn't add them all in another 6 hours. It shows you the name, and how many scenes the performer is in and if the site currently exists in your library. The user can update this list as needed to add and remove options on the fly. 

Note for other devs here: SKYHOOK NEEDS UPDATED FIRST. I HAVE IT ALREADY JUST NEED TO PR

Button:
8ac539c4ade139bf7517a4fa9e55d28e

Modal: 

3b6d8a55df9c6c37d0766a300bf12849

Note: removing an option that was added previously WILL NOT remove it from their instance. They still need to manually delete it


sampulsar (approved):

JoyboySon:
There appears to be a bug, unless intentional. If a site is already in the exclusion list, it doesn't show the site's name; instead, it shows this: 
image

Gykes:
> There appears to be a bug, unless intentional. If a site is already in the exclusion list, it doesn't show the site's name; instead, it shows this: image

This is due to skyhook not being updated. My images use a local version that I use to develop off of. Once the cache server is updated then yours will work accordingly. I have already PR'd the update in the server and just waiting for deployment

I appreciate the initiative with testing, I really didn't think people cared lol

PR #779, opened 2025-12-15
New: HHD800 Parsing

Gykes:
Add support for HHD800 parsing for JAV. They seem to be one of the larger JAV groups


sampulsar (commented):
You should write that example as a unit test.

sampulsar (approved):

PR #783, opened 2025-12-15
New: Trim Scene Title when file name exceeds the max file length

sampulsar:
#### Database Migration
NO

#### Description
Make sure the the filename is within the limit allowed for a file system.

Remove a Character from the title until the filename is under the limit 

#### Todos
- [x] Tests

* Fixes #XXXX


Gykes (commented):
Would it not be a little easier to just trunc the title token similar to how I did it on V2? This way probably works are well just seems more complex.

Gykes:
Nevermind i actually found a usecase where yours. Specifically for JAV. Will more than likely implement something similar in V2

Gykes (approved):

PR #789, opened 2025-12-17
Fix: Refresh Studios and Performers oldest to newest

sampulsar:
#### Database Migration
NO

#### Description
Refresh the Studios and Performers based on last sync so older ones are synced first. The previous alpha order means that if a studio causes a fault then the end of the alphabet may never get a refresh. 

#### Todos
- [x] Tests

* Fixes #XXXX


Gykes:
I believe this is actually a native Arr issue. Things like Huntarr were created as, apparently, when the refresh task kicks off it just stops at a certain point. Depending on how many sites you have it can essentially never refresh after a certain letter as it restarts when the next refresh kicks in.

Will do this for V2 as well.

Gykes (approved):

PR #826, opened 2025-12-25
New: Auto Rename Folder and File

sampulsar:
#### Database Migration
NO

#### Description
Complete some functionality so that a scene/movie will rename when a scan is triggered.

Some data can change like the titles within https://stashdb.org/studios/0342d621-df03-4624-9528-63755abcfc13

Also allows for the file and folder format to be changed and then applied on the next refresh.

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -134,6 +134,8 @@
   "AutoRedownloadFailedFromInteractiveSearch": "Redownload Failed from Interactive Search",
   "AutoRedownloadFailedFromInteractiveSearchHelpText": "Automatically search for and attempt to download a different release when failed release was grabbed from interactive search",
   "AutoRedownloadFailedHelpText": "Automatically search for and attempt to download a different release",
+  "AutoRenameFolders": "Automatically Rename Folders and Files",
+  "AutoRenameFoldersHelpText": "A rename event will be triggered daily when the movie is refreshed. WARNING: This should not be enabled if any external tool is used to rename files in Whisparr's folders.",
   "AutoTagging": "Auto Tagging",
   "AutoTaggingLoadError": "Unable to load auto tagging",
   "AutoTaggingNegateHelpText": "If checked, the auto tagging rule will not apply if this {implementationName} condition matches.",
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -134,6 +134,8 @@
   "AutoRedownloadFailedFromInteractiveSearch": "Redownload Failed from Interactive Search",
   "AutoRedownloadFailedFromInteractiveSearchHelpText": "Automatically search for and attempt to download a different release when failed release was grabbed from interactive search",
   "AutoRedownloadFailedHelpText": "Automatically search for and attempt to download a different release",
+  "AutoRenameFolders": "Automatically Rename Folders and Files",
+  "AutoRenameFoldersHelpText": "A rename event will be triggered when a movie is refreshed",
```
plz12345: @sampulsar 
This should probably have an explicit warning not to be used if another process is renaming files outside of Whisparr.  I know we discourage it, but I can see the Discord complaints happening on this one, since we have a daily refresh happening on all movies/scenes, aside from this new setting that that daily refresh would trigger.  Basically  a mass-rename, every day, if mis-configured.

src/NzbDrone.Core/Movies/RefreshMovieService.cs
```diff
@@ -190,6 +197,22 @@ private Movie RefreshMovieInfo(int movieId)
...
+                    {
+                        _logger.Info("Renaming movie folder for {0} to {1}", movie, renameMovieFilePreview.NewPath);
+                        files.Add(renameMovieFilePreview.MovieFileId);
+                    }
+
+                    _commandQueueManager.Push(new RenameFilesCommand(movie.Id, files));
+                }
+            }
 
             _logger.Debug("Finished movie metadata refresh for {0}", movieMetadata.Title);
```
plz12345, line 216: You have `Info` for the logging in the previous loop, but `Debug` here.  I think these levels should align, so the process logs properly.  Or reverse them, so the verbose logging is on Debug, not Info.

plz12345 (changes requested):

src/NzbDrone.Core/Localization/Core/en.json
sampulsar: Updating to "A rename event will be triggered daily when the movie is refreshed. WARNING: This should not be enabled if any external tool is used to rename files in Whisparr's folders."

src/NzbDrone.Core/Movies/RefreshMovieService.cs
sampulsar: Remove the log for rename as the renameCommand logs the information.

plz12345 (approved):

PR #868, opened 2025-12-31
Fix: Movie import

sampulsar:
#### Database Migration
NO

#### Description
Fix the movie import so that only movies are the candidates, stops issues where the movie is also released as scenes with the same title.

#### Todos
- [x] Tests

* Fixes #XXXX


plz12345 (approved):
It looks like this is only called from methods specific to Movie.ItemType == Movie, so no concerns.

PR #871, opened 2025-12-31
Fix: Show Movie Performers

sampulsar:
#### Database Migration
NO

#### Description
Load Whisparr before the movies, performers, studios are populated.

Show all cast if returned from skyhook, allows for a future skyhook change to return non-mapped performer information as cast.

Only Add Performers that have a StashId.

Fix the Performers page listing studios that are not linked in StashDB.

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Helpers/Hooks/useAppPage.ts
+++ b/frontend/src/Helpers/Hooks/useAppPage.ts
@@ -88,10 +88,7 @@ const useAppPage = () => {
 
   const isPopulated = useSelector(
     (state: AppState) =>
-      state.movies.isPopulated &&
       state.customFilters.isPopulated &&
-      state.performers.isPopulated &&
-      state.studios.isPopulated &&
       state.tags.isPopulated &&
       state.settings.ui.isPopulated &&
       state.settings.qualityProfiles.isPopulated &&
--- a/frontend/src/Movie/Details/Credits/Cast/MovieCastPoster.tsx
+++ b/frontend/src/Movie/Details/Credits/Cast/MovieCastPoster.tsx
@@ -40,22 +40,24 @@ function MovieCastPoster({
   } as React.CSSProperties;
 
   const contentStyle = { width: `${posterWidth}px` } as React.CSSProperties;
-
-  if (!performer?.foreignId) return null;
-
-  const link = `/performer/${performer.foreignId}`;
+  const name = performer.fullName ?? performer.name;
+  const isPerformer = !!performer?.foreignId;
+  const isPerfotmerLoaded = !!performer?.fullName;
+  const link = isPerformer ? `/performer/${performer.foreignId}` : '';
 
   return (
     <div className={styles.content} style={contentStyle}>
       <div className={styles.posterContainer}>
-        <div className={styles.controls}>
-          <MonitorToggleButton
-            className={styles.action}
-            monitored={performer.monitored}
-            size={20}
-            onPress={onTogglePerformerMonitored}
-          />
-        </div>
+        {isPerfotmerLoaded && (
+          <div className={styles.controls}>
+            <MonitorToggleButton
+              className={styles.action}
+              monitored={performer.monitored}
+              size={20}
+              onPress={onTogglePerformerMonitored}
+            />
+          </div>
+        )}
 
         <div style={elementStyle}>
           <Link className={styles.link} to={link}>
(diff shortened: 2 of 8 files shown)


frontend/src/Performer/Details/PerformerDetailsStudio.js
```diff
@@ -181,7 +181,8 @@ class PerformerDetailsStudio extends Component {
     } = getStudioStatistics(items);
 
     const sizeOnDisk = _.sumBy(items, 'sizeOnDisk');
-    const studioLink = `/studio/${foreignId}`;
+    const isStudio = foreignId !== undefined;
```
plz12345: @sampulsar 
`foreignId != null` would check for both null and undefined, safer here.  Or !!foreignId if you just want to assert it has any value.

```
foreignId !== undefined - checks only for undefined. It treats null, 0, '', false as valid values.
!!foreignId -  coerces to boolean and checks truthiness. It returns false for undefined, null, 0, '', false, NaN.
```

frontend/src/Performer/Details/PerformerDetailsStudioConnector.js
```diff
@@ -85,6 +86,12 @@ function createMapStateToProps() {
       // Sort once filtered
       items = sort(items, performerScenes);
 
+      if (studio?.title === undefined) {
```
plz12345: @sampulsar this is safer here.  Validates existing studio object, checks studio.title against null, undefined, and empty strings, then proceeds with your pseudo studio object creation if needed.  CHecking against just undefined is fragile.

```
if (!studio || studio.title == null || studio.title.trim() === '') {
  studio = studio ?? {};
  studio.id = 0;
  studio.title = translate('NoStashDBStudioLink');
  studio.monitored = false;
}
```

frontend/src/Movie/Details/Credits/Cast/MovieCastPoster.tsx
```diff
@@ -40,22 +40,24 @@ function MovieCastPoster({
...
       <div className={styles.posterContainer}>
-        <div className={styles.controls}>
-          <MonitorToggleButton
-            className={styles.action}
-            monitored={performer.monitored}
-            size={20}
-            onPress={onTogglePerformerMonitored}
-          />
-        </div>
+        {isPerfotmerLoaded && (
```
plz12345, line 51: @sampulsar 
Can you use `isPerformer` instead, here?  The intent is that we don't even show the toggle if the performer isn't StashDB, right?  There's nothing to monitor for pseudo performers you are showing purely based on TPDB data, I assume.

frontend/src/Helpers/Hooks/useAppPage.ts
```diff
@@ -88,10 +88,7 @@ const useAppPage = () => {
 
   const isPopulated = useSelector(
     (state: AppState) =>
-      state.movies.isPopulated &&
```
plz12345, line 91: @sampulsar was removing these 3 just to make the app appear to load faster?  Is it a dependency for this PR, or just added on?  I know they per-object hooks are called anyways by their respective top-level components, so I assume this is just "load the UI faster, do the rest after".  Please confirm.

> Load Whisparr before the movies, performers, studios are populated.

frontend/src/Movie/Details/Credits/Cast/MovieCastPoster.tsx
```diff
@@ -40,22 +40,24 @@ function MovieCastPoster({
   } as React.CSSProperties;
 
   const contentStyle = { width: `${posterWidth}px` } as React.CSSProperties;
-
-  if (!performer?.foreignId) return null;
-
-  const link = `/performer/${performer.foreignId}`;
+  const name = performer.fullName ?? performer.name;
+  const isPerformer = performer?.foreignId !== '';
+  const isPerfotmerLoaded = performer.fullName !== undefined;
```
plz12345: @sampulsar  , safer version

```
const isPerformer = !!performer?.foreignId; 
const isPerfotmerLoaded = !!performer?.fullName;
```
 , assuming foreignId and fullName are never just whitespace.  I know foreignId is non-nullable on the object, but didn't look at fullName.

I"m being picky here because of the efforts that I put into converting these to TypeScript.

plz12345 (changes requested):

frontend/src/Helpers/Hooks/useAppPage.ts
sampulsar: This loads the UI faster as it was added in a Radarr merge

frontend/src/Movie/Details/Credits/Cast/MovieCastPoster.tsx
sampulsar: isPerformerLoaded will be false also if the performer state has not been populated.

plz12345 (approved):

PR #874, opened 2026-01-01
Fix: Correct the Refresh of Movies to only run when moviesMonitored is enabled.

sampulsar:
#### Database Migration
NO

#### Description
Make sure that movieMonitored is enabled when performers are refreshed.

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/Movies/Performers/RefreshPerformerService.cs
+++ b/src/NzbDrone.Core/Movies/Performers/RefreshPerformerService.cs
@@ -175,7 +175,7 @@ namespace NzbDrone.Core.Movies.Performers
                 _logger.Info("Synced performer {0} has {1} scenes adding {2} and added {3}", performer.Name, performerWork.Scenes.Count, scenesToAdd.Count(), scenesAdded);
             }
 
-            if (performer.Monitored)
+            if (performer.MoviesMonitored)
             {
                 var moviesAdded = 0;
 


src/NzbDrone.Core/Movies/Performers/RefreshPerformerService.cs
```diff
@@ -175,7 +175,7 @@ private void SyncPerformerItems(Performer performer)
                 _logger.Info("Synced performer {0} has {1} scenes adding {2} and added {3}", performer.Name, performerWork.Scenes.Count, scenesToAdd.Count(), scenesAdded);
             }
 
-            if (performer.Monitored)
+            if (performer.MoviesMonitored)
```
plz12345, line 178: @sampulsar 
What is the functional difference between these two props?  They same to do the same thing and also seem to be in sync with each other, at least when toggling this on a performer.

sampulsar: Under the edit of the performer, you will see MovieMonitored. 

I a future PR the UI will be updated so that they both display on Performers and Studios. Only the Monitored option is displayed that is untended to be used for just scenes, as that was what it has been used for as there were limited movies.

Finally the MovieMetaProvider setting will be used to hide the movie monitoring and section depending on the settings.

plz12345: Oh, so one will be for scenes and one for movies?

sampulsar: Yes, allow you to control what you monitor, as you may not want to use Whisparr for movies.

sampulsar: The UI side will take a while so it is a separate PR. This can spot random movies showing for users as the TMDB ids are populated.

plz12345 (dismissed):

plz12345 (approved):

PR #903, opened 2026-01-06
Fix: Do not map the Studio from Sky Hook if it is not mapped

sampulsar:
#### Database Migration
NO

#### Description
Map the Movie Studio correctly for TPDB, and pending TMDB

Client would add the studio for TPDB but the link would initially be incorrect in the client when navigating from the add form directly after the add. The movie refresh that is triggered at the add correct the link by may be incorrect for a second. It believed that the tpdbid for the studio was the StashDB id.

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
+++ b/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
@@ -183,7 +183,7 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
 
             var performers = httpResponse.Resource.Credits.Select(c => MapPerformer(c.Performer)).DistinctBy(p => p.ForeignId).ToList();
 
-            return new Tuple<MovieMetadata, Studio, List<Performer>>(movie, null, performers);
+            return new Tuple<MovieMetadata, Studio, List<Performer>>(movie, MapStudio(httpResponse.Resource.Studio), performers);
         }
 
         public Tuple<MovieMetadata, Studio, List<Performer>> GetTpdbMovieInfo(string tpdbId)
@@ -536,13 +536,11 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
             }
             else if (resource.Studio.ForeignIds != null && resource.Studio.ForeignIds.TmdbId > 0)
             {
-                movie.StudioForeignId = resource.Studio.ForeignIds.TmdbId.ToString();
                 movie.StudioTitle = resource.Studio.Title;
                 movie.Studio = resource.Studio;
             }
             else if (resource.Studio.ForeignIds != null && resource.Studio.ForeignIds.TpdbId.IsNotNullOrWhiteSpace())
             {
-                movie.StudioForeignId = resource.Studio.ForeignIds.TpdbId;
                 movie.StudioTitle = resource.Studio.Title;
                 movie.Studio = resource.Studio;
             }
@@ -1350,6 +1348,12 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
 
         private Studio MapStudio(StudioResource studio)
         {
+            // Do not Map the Studio if it has not mapped to a StashDB studio. For Movies as the name is stored in the movie Metadata.
+            if (string.IsNullOrEmpty(studio?.ForeignIds?.StashId))
+            {
+                return null;
+            }
+
             var newStudio = new Studio
             {
                 Title = studio.Title,
--- a/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
+++ b/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
@@ -394,7 +394,7 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
             return MapStudio(httpResponse.Resource);
         }
 
-        public (List<string> Scenes, List<string> TpdbMovies, List<int> Movies) GetPerformerWorks(string stashId)
+        public (List<string> StashdbIds, List<string> TpdbIds, List<int> TmdbIds) GetPerformerWorks(string stashId)
         {
             var httpRequest = _whisparrMetadata.Create()
                                              .SetSegment("route", "performer")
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
```diff
@@ -452,7 +452,7 @@ public List<string> GetStudioScenes(string stashId)
             return scenes;
         }
 
-        public (List<string> Scenes, List<string> TpdbMovies) GetStudioWorks(string stashId)
+        public (List<string> Scenes, List<string> TpdbMovies, List<int> Movies) GetStudioWorks(string stashId)
```
plz12345: @sampulsar would you mind renaming these vars to be more accurate to what they are?  Assuming

- stashdbIds
- tpdbIds
- tmdbIds

makes the code easier to follow when reading later down the road.

src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
plz12345: Suggesting a refactor on this chain of if/else.  Only needs to nullcheck once per risky assignment, easier to read, same results.

```csharp
var foreignIds = resource.Studio?.ForeignIds;
if (foreignIds != null)
{
    if (!string.IsNullOrWhiteSpace(foreignIds.StashId))
    {
        movie.StudioForeignId = foreignIds.StashId;
        movie.StudioTitle = resource.Studio.Title;
        movie.Studio = resource.Studio;
    }
    else if (foreignIds.TmdbId > 0 || !string.IsNullOrWhiteSpace(foreignIds.TpdbId))
    {
        movie.StudioTitle = resource.Studio.Title;
        movie.Studio = resource.Studio;
    }
}
```

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
```diff
@@ -138,12 +137,46 @@ private void SyncStudioItems(Studio studio)
...
             }
 
             if (studio.MoviesMonitored)
             {
+                if (_configService.WhisparrMovieMetadataSource == MovieMetadataType.TMDB)
+                {
+                    var moviesAdded = 0;
+                    var tmbdId = 0;
+                    var existingMovies = _movieService.AllMovieTmdbIds();
+                    var excludedMovies = _importListExclusionService.GetAllExclusions().Select(e => int.TryParse(e.ForeignId, out tmbdId)).Select(e => tmbdId).Where(e => e != 0).ToList();
```
plz12345: @sampulsar Can you reformat for readability?

```csharp
var excludedMovies = _importListExclusionService.GetAllExclusions()
    .Select(e => int.TryParse(e.ForeignId, out tmbdId))
    .Select(e => tmbdId).Where(e => e != 0)
    .ToList();
```

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
```diff
@@ -138,12 +137,46 @@ private void SyncStudioItems(Studio studio)
...
+                    var moviesToAdd = studioWork.Movies.Where(m => !existingMovies.Contains(m)).Where(m => !excludedMovies.Contains(m));
```
plz12345: @sampulsar please format for readability and performance?  You can avoid the chaining and repeated scan of studioWork.Movies (I don't know how big that'd be, but assuming could be hundreds+)

```csharp
var moviesToAdd = studioWork.Movies
    .Where(m => !existingMovies.Contains(m) && !excludedMovies.Contains(m));
```

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
```diff
@@ -138,12 +137,46 @@ private void SyncStudioItems(Studio studio)
...
+
+                    if (moviesToAdd.Any())
+                    {
+                        var movieLists = moviesToAdd.Select(m => new Movie
+                        {
+                            ForeignId = m.ToString(),
```
plz12345, line 167: @sampulsar what happens when someone adds a TMDB Studio URL to a StashDB studio?  Does this ForeignId get updated?  I know StashDB doesn't has that as an "allowed" URL type, but we probably should request it to futuyre proof this.

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
```diff
@@ -138,12 +137,46 @@ private void SyncStudioItems(Studio studio)
...
+                            TmdbId = m,
+                            QualityProfileId = studio.QualityProfileId,
+                            RootFolderPath = studio.RootFolderPath,
+                            AddOptions = new AddMovieOptions
+                            {
+                                SearchForMovie = studio.SearchOnAdd,
+                                AddMethod = AddMovieMethod.Studio
+                            },
+                            Monitored = true,
```
plz12345, line 176: @sampulsar should this be hard-coded to true?  This var is scene monitoring, no?

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
```diff
@@ -138,12 +137,46 @@ private void SyncStudioItems(Studio studio)
...
+                            Tags = studio.Tags
+                        }).Chunk(chunkSize);
+
+                        foreach (var movieList in movieLists)
+                        {
+                            moviesAdded += _addMovieService.AddMovies(movieList.ToList(), true).Count;
+                        }
+                    }
+
+                    _logger.Info("Synced studio {0} has {1} movies adding {2} and added {3}", studio.Title, studioWork.Movies.Count, moviesToAdd.Count(), moviesAdded);
```
plz12345: @sampulsar pls format for readability.  I know the code is lousy with line-wrapping log statements, but we can fix what we touch/add :)

```csharp
_logger.Info(
    "Synced studio {0} has {1} movies adding {2} and added {3}",
    studio.Title,
    studioWork.Movies.Count,
    moviesToAdd.Count(),
    moviesAdded
);
```

plz12345 (changes requested):

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
sampulsar: Skyhook will process the link and add the mapping on the server. The change to the studio in skyhook causes an invalidation of the studio works, and also invalidates all of the movies for that TMDB studio.

The refresh that happen on the users end will pull new movie data where the StashId in ForeignIds is populated.

I have been testing this for a few days with TPDB, as I have been submitting performer and studio links into StashDB.

I have requested that the regex for TMDB be altered to allow for the links to the TMDB studios (company).

src/NzbDrone.Core/Movies/Studios/RefreshStudioService.cs
sampulsar: If you are monitoring a Studio then all of the new items that are synced are set to monitored. 

If the studio is not monitored then this code does not run as the studio's scenes, or movies are not added in an unmonitored stated.

If is did it for all unmonitored, then everyone whisparr instance would have every performer, studio, movie that exists.

sampulsar:
@plz12345 

Update or answered.

plz12345 (approved):

PR #912, opened 2026-01-09
New: API Docs Icon

JoyboySon:
#### Database Migration
NO

#### Description
Proposal/suggestion for button in API settings for easier access and visibility to the SwaggerUI API page. 

image

#### Todos
- [ ] Tests
- [ ] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)

N/A


JoyboySon:
Would it be better to put a link in System > Status > More Info instead?

plz12345:
I'm thinking as a "Docs" button here, next to the red recycle button.  Suggest blue similar to the ones on the media management page with the "?" buttons next to the naming textboxes.  Also, please add a screenshot since this is UI related.

image

plz12345:
instead of the ASCII ? , try icons.UNKNOWN

Something like (I didn't test):

```typescript
<FormInputButton
  key="apidocumentation"
  kind={kinds.DEFAULT}
  to="/docs"
>
  <Icon
    name={icons.UNKOWN}
  />
</FormInputButton>
```

We render that as Font Awesome's "question" icon.

image

JoyboySon:
image

Fixed. Also, does it look a little off-center?

plz12345:
Maybe by a pixel or few, but not crazy.  It's likely do to the effect that is rounding the outside edge of the button.  It's acceptable, IMO.  You could add CSS to the button itself and try bumping over  children (the icon).  I don't think it's a big deal.  Plenty of other fun to be had.

JoyboySon:
K. I'll just leave it as it is then.

plz12345 (approved):

PR #915, opened 2026-01-10
New: Add Token without SE in Scene Title

JoyboySon:
#### Database Migration
NO

#### Description
{Scene CleanTitle} includes the season and episode numbers, but some media players don't play well with that, so I implemented a new token like Scene CleanTitle but removes the "S#:E#" as {Scene CleanTitleNoSeasonEpisode}.

image

#### Todos
- [x] Tests

N/A

### DIFF
--- a/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs
+++ b/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs
@@ -202,6 +202,46 @@ namespace NzbDrone.Core.Test.OrganizerTests.FileNameBuilderTests
                    .Should().Be("Pure Taboo - 2025-11-25 - The Last Train Home [Maddy OReilly Reagan Foxx Rissa May]");
         }
 
+        [TestCase("My Scene Title S01:E02", "My Scene Title")]
+        [TestCase("My Scene Title s01:e02", "My Scene Title")]
+        [TestCase("My Scene Title S01:e02", "My Scene Title")]
+        [TestCase("My Scene Title s01:E02", "My Scene Title")]
+        [TestCase("My Scene Title S1:E2", "My Scene Title")]
+        [TestCase("My Scene Title S1:E02", "My Scene Title")]
+        [TestCase("My Scene Title S01:E2", "My Scene Title")]
+        [Test]
+        public void scene_clean_title_no_se_Title(string title, string expected)
+        {
+            _scene.Title = title;
+            _namingConfig.StandardSceneFormat = "{Scene CleanTitleNoSeasonEpisode}";
+
+            Subject.BuildFileName(_scene, _movieFile)
+                .Should().Be(expected);
+        }
+
+        [Test]
+        public void scene_clean_title_no_se_format()
+        {
+            _scene.Title = "My Scene Title S01:E02";
+            _scene.MovieMetadata.Value.ReleaseDate = "2025-11-25";
+
+            _namingConfig.StandardSceneFormat = "{Studio Title} - {Release-Date} - {Scene CleanTitleNoSeasonEpisode}";
+
+            Subject.BuildFileName(_scene, _movieFile)
+                .Should().Be("Pure Taboo - 2025-11-25 - My Scene Title");
+        }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


JoyboySon:
Not sure what to put for token's example.

plz12345:
Yeah, that's a tough one, since you're displaying a negative.  Maybe instead of "NoSE, do "NoSeason".  You might even be able to fit NoSeasonEpisode in there without breaking to a new line (or breaking CSS...).

Also, will need a unit test for this in `src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs`

sampulsar (commented):
Covers specifically S31:E1 for a site like Step Siblings Caught

https://stashdb.org/studios/3b88b263-ac4f-4fc6-8031-38daa05d452d

src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs
```diff
@@ -202,6 +202,39 @@ public void scene_clean_female_performers_format()
                    .Should().Be("Pure Taboo - 2025-11-25 - The Last Train Home [Maddy OReilly Reagan Foxx Rissa May]");
         }
 
+        [Test]
+        public void scene_clean_title_no_se_Title()
```
sampulsar: The test could be alter to show exact change, if this is what is expected 

        [TestCase("My Scene Title S01:E02", "My Scene Title")]
        [TestCase("My Scene Title S01 E02", "My Scene Title S01 E02")]
        [TestCase("My Scene Title S01E02", "My Scene Title S01E02")]
        [Test]
        public void scene_clean_title_no_se_Title(string title, string expected)
        {
            _scene.Title = title;
            _namingConfig.StandardSceneFormat = "{Scene CleanTitleNoSeasonEpisode}";

            Subject.BuildFileName(_scene, _movieFile)
                .Should().Be(expected);
        }

plz12345:
It's not specific to that studio, but as long as the regex can cover nuance like:

- S1:E1 - caps, single digit for both
- S24:E1 - mixed single and double digits
- S10:e09 - mixed case, double digit for both
- s12:e14 - lower case for season and episode
- s09:e10 - lowercase with double digits

Should be OK.  I think those variants should be added to the unit tests to be sure.

sampulsar:
It currently only covers when ":" is the delimiter between the Season and Episode, and they are capitals.

The test should cover all

The following are not covered by the change.
My Scene Title s01:e02
My Scene Title S01 E02
My Scene Title S01E02

JoyboySon:
I tested and it does ignore case. Updated the tests to reflect that. 

As for "S01 E02" & "S01E02", I think stashdb follows the "S#:E#" rule pretty strictly, so I haven't come across those different versions.

sampulsar:
You should alter the test so that the one test has multiple test cases to cover both instead of hard setting any values.

The test cases are then the documentation of what the function needs to do, so that another person does not alter the function and breaks the intended behaviour. 

It can also be useful for any edge cases that should not change the title.

    [TestCase("My Scene Title S01:E02", "My Scene Title")]
    [TestCase("My Scene Title s01:e02", "My Scene Title")]
    [Test]
    public void scene_clean_title_no_se_Title(string title, string expected)
    {
        _scene.Title = title;
        _namingConfig.StandardSceneFormat = "{Scene CleanTitleNoSeasonEpisode}";

        Subject.BuildFileName(_scene, _movieFile)
            .Should().Be(expected);
    }

JoyboySon:
@sampulsar 
So like this?

```
        [TestCase("My Scene Title S01:E02", "My Scene Title")]
        [TestCase("My Scene Title s1:e2", "My Scene Title")]
        [TestCase("My Scene Title S1:e02", "My Scene Title")]
        [Test]
        public void scene_clean_title_no_se_Title(string title, string expected)
        {
            _scene.Title = title;
            _namingConfig.StandardSceneFormat = "{Scene CleanTitleNoSeasonEpisode}";

            Subject.BuildFileName(_scene, _movieFile)
                .Should().Be(expected);
        }
```
Edit: added a few more test cases in the new commit

sampulsar (approved):

sampulsar:
@JoyboySon are you able to squash your commits into one?

PR #919, opened 2026-01-11
New: add Alternate Titles

sampulsar:
#### Database Migration
NO

#### Description
Add the Alternate Title functionality from Radarr back into Whisparr.

The Index search now also uses the Alternate titles.

Alternate Titles will be extracted via the following with some alternate names, eg removing the vol etc.
https://github.com/Whisparr/WhisparrAPI.Metadata/pull/34

There is a null check on the skyhook data so that it works while the data in transitioned.

image

#### Todos
- [x] Tests

* Fixes #XXXX


plz12345:
just to confirm, this won't work until that skyhook change is live, but won't break?  I can build and run locally, but just want to be sure.

sampulsar:
There will not be any alternate title data until, the scene/movie has been repopulated once skyhook has been updated.

The null check in SkyHookProxy.cs‎ allows it to work currently without any data as the missing property is then treaded as an empty list.

plz12345 (approved):

PR #941, opened 2026-01-16
Fix: Regex Pattern for ReleaseGroupRegex

CuddleBear92:
adding support for .ReleaseGroup in the regex parser. This tiny change should not break anything cause its just an optional lookup in the regex and will add for .ReleaseGroup for example .WRB, .KTR, .PRT that can happen.

Cause this was not supported before, matching anything with a . as a release group would just make it fail and not let it be found and it will not apply the custom format as a release group.

Please make any changes if needed. Have not tested it but i see no reason why it shouldn't just work fine.

Did not make a github issue on it yet! mentioned it earlier [today on discord](https://discord.com/channels/957516730727563285/1461555604748505362/1461771902220632249) and before on discord way back.

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -221,7 +221,7 @@ namespace NzbDrone.Core.Parser
         private static readonly Regex CleanQualityBracketsRegex = new Regex(@"\[[a-z0-9 ._-]+\]$",
                                                                    RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex ReleaseGroupRegex = new Regex(@"(-|.)(?<releasegroup>[a-z0-9]+(?<part2>-[a-z0-9]+)?(?!.+?(?:480p|576p|720p|1080p|2160p)))(?<!(?:WEB-DL|Blu-Ray|480p|576p|720p|1080p|2160p|DTS-HD|DTS-X|DTS-MA|DTS-ES|-ES|-EN|-CAT|[ ._]\d{4}-\d{2}|-\d{2})(?:\k<part2>)?)(?:\b|[-._ ]|$)|[-._ ]\[(?<releasegroup>[a-z0-9]+)\]$",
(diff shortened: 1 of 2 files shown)


Gykes:
So, I think it would be better to modify it to `[.-]` I think with the `.` being outside brackets it becomes special and it will just match anything. Inside brackets it becomes literal and will match a `.`

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -221,7 +158,7 @@ public static class Parser
         private static readonly Regex CleanQualityBracketsRegex = new Regex(@"\[[a-z0-9 ._-]+\]$",
                                                                    RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex ReleaseGroupRegex = new Regex(@"-(?<releasegroup>[a-z0-9]+(?<part2>-[a-z0-9]+)?(?!.+?(?:480p|576p|720p|1080p|2160p)))(?<!(?:WEB-DL|Blu-Ray|480p|576p|720p|1080p|2160p|DTS-HD|DTS-X|DTS-MA|DTS-ES|-ES|-EN|-CAT|[ ._]\d{4}-\d{2}|-\d{2})(?:\k<part2>)?)(?:\b|[-._ ]|$)|[-._ ]\[(?<releasegroup>[a-z0-9]+)\]$",
+        private static readonly Regex ReleaseGroupRegex = new Regex(@"(-|.)(?<releasegroup>[a-z0-9]+(?<part2>-[a-z0-9]+)?(?!.+?(?:480p|576p|720p|1080p|2160p)))(?<!(?:WEB-DL|Blu-Ray|480p|576p|720p|1080p|2160p|DTS-HD|DTS-X|DTS-MA|DTS-ES|-ES|-EN|-CAT|[ ._]\d{4}-\d{2}|-\d{2})(?:\k<part2>)?)(?:\b|[-._ ]|$)|[-._ ]\[(?<releasegroup>[a-z0-9]+)\]$",
```
Gykes:
```suggestion
        private static readonly Regex ReleaseGroupRegex = new Regex(@"[.-](?<releasegroup>[a-z0-9]+(?<part2>-[a-z0-9]+)?(?!.+?(?:480p|576p|720p|1080p|2160p)))(?<!(?:WEB-DL|Blu-Ray|480p|576p|720p|1080p|2160p|DTS-HD|DTS-X|DTS-MA|DTS-ES|-ES|-EN|-CAT|[ ._]\d{4}-\d{2}|-\d{2})(?:\k<part2>)?)(?:\b|[-._ ]|$)|[-._ ]\[(?<releasegroup>[a-z0-9]+)\]$",
```

Gykes (dismissed):

Gykes:
Will also need to update unit tests to handle this regex change.

Gykes:
Just so you are aware I went a different route. I cannot stand regex and every route I took I would solve one problem and create another. I kept the original regex the same and added a fallback to handle your use case. 

I looked at the code for how this will update existing scenes and you're not going to like it. The refresh and rescan will only reparse on new files it seems like. All new files should have the release group applied but current ones will need to be removed and re imported or manually added.

Gykes (approved):
This is a longer more robust way of handing the situation. Is it ideal? Probably not. The problem with a single regex to handle everything is there are always things that slips through the cracks. My current implementation is more robust and catches extra formats

PR #945, opened 2026-01-17
New: Series Type for JAV

### LINKED ISSUE #944: [V2] Expand JAV to be More Like Anime in Sonarr

Develop (V2 "pink")

Currently, there is no way to distinguish JAV from scenes. This causes issues with naming as JV realistically should only need the Studio code for the filename. Currently this is unsupported as only one profile (Scenes) is in use. Users are having naming issues as JAV naming is..... wack

Similar to Anime from Sonarr. Allow for separate profiles for JAV and scenes with independent naming requirements

Nah

### Anything else?

_No response_

Gykes:
This PR adds JAV-specific naming support by introducing a Series Type field that allows users to distinguish JAV sites from regular sites. When a series is set to the "JAV" type, files will be renamed using a separate "JAV Episode Format" instead of the standard episode format. This follows the same pattern Sonarr uses for Anime series types. The feature includes a new `{External Id}` naming token which outputs the episode's external identifier (e.g., "SSIS-123"), making it easy to name JAV files by their Studio ID. Users can set the series type when adding a new series or edit it on existing series, and configure the JAV naming format in `Settings > Media Management > Naming`. This addresses the need for different naming conventions between western adult content (which typically uses site/date/title format) and JAV content (which is commonly organized by catalog number). This will more than likely be more changes where the next steps are to allow JAV specific searches and remove it from the indexer advanced settings

Will be requesting that JAV Users test this before I merge it in. 

Naming Screenshot:
Namingsetting

Add New/Edit Modal:
UImodal

Fixes: https://github.com/Whisparr/Whisparr/issues/944


plz12345 (approved):
I reviewed code only, and looks solid.  Didn't notice any glaring issues, so if build and run is clean, no concerns.

PR #950, opened 2026-01-19
Chore: Remove Unique Requirement for Unit Tests

Gykes:
Dumb issue where >1 exact scene causes certain unit tests to fail. Removing the check for uniqueness .


Gykes:
Once this gets merged I will rebase the other PR so it passes tests

PR #996, opened 2026-02-03
Fix: Handle null episode in manual interaction notifications

Gykes:
Was getting various errors in logs like so 

```
[Error] EventAggregator: NotificationService failed while processing [ManualInteractionRequiredEvent]

  [v2.1.0.63] System.NullReferenceException: Object reference not set to an instance of an object.
     at NzbDrone.Core.Notifications.NotificationService.Handle(ManualInteractionRequiredEvent message) in ./Whisparr.Core/Notifications/NotificationService.cs:line 217
     at NzbDrone.Core.Messaging.Events.EventAggregator.PublishEvent[TEvent](TEvent event)
```

This was happening due to non matches episodes in the queue. The notification call was trying to run but the episode data was null due to no match. Now it returns early and logs to prevent the null reference

### DIFF
--- a/src/NzbDrone.Core/Notifications/NotificationService.cs
+++ b/src/NzbDrone.Core/Notifications/NotificationService.cs
@@ -214,6 +214,12 @@ namespace NzbDrone.Core.Notifications
 
         public void Handle(ManualInteractionRequiredEvent message)
         {
+            if (message.Episode == null)
+            {
+                _logger.Trace("Skipping manual interaction notification, episode data is not available");
+                return;
+            }
+
             var manualInteractionMessage = new ManualInteractionRequiredMessage
             {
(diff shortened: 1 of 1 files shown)


plz12345:
Is the "no mach in queue" part due to someone trying to add a site rip or something?  I thought when you did an override, it added that scene data to the report.

Gykes:
So, mostly this comes from if someone added an external scene or siterip from outside of Whisparr and there is no history. 

Your comment did get me thinking, We still want to send off Manual Interact notification even if null so I modified it to use the download title as the message.

plz12345 (approved):
This looks good.  I would recommend revising the PR subject line to reflect the change from simple null check to something more relevant.

PR #1026, opened 2026-02-25
New: Make JAV Tab

Gykes:
This is a weird one that I've been thinking about for a while. Currently all sites, JAV and "regular" get clustered into one large page and makes it difficult to parse without using tags, filtering, and all that other jazz. This breaks it out based on Series.Type. When you select the JAV series type it will move that site to the JAV tab and opposite for Scenes. All currently existing sites with the type will automatically move to the appropriate page. 

The only things that I see currently being an issue are import lists and library import. Currently both of those default to "scenes" so they will go there first and you will need to manually modify the series type to be JAV. That is an issue I beleive is better tackled in a separate PR as this one is already quite large.

Will be requesting testing before having this merged
UI: 
62635507b5c3f3b00c68714b9095c0e5


Gykes:
It says builds fail but its that stupid SonarCube thing.

sampulsar (approved):

PR #1034, opened 2026-02-27
fix: undefined error on delete modal for undefined series.statistics …

Venipa:
…prop - fallback to 0 values

#### Database Migration
NO

#### Description
I was trying to delete multiple shows that got accidentally added through import list and without refreshing them to fill the statistics property this error occurred.

image
image

### DIFF
--- a/frontend/src/Series/Index/Table/SeriesIndexRow.tsx
+++ b/frontend/src/Series/Index/Table/SeriesIndexRow.tsx
@@ -324,7 +324,7 @@ function SeriesIndexRow(props: SeriesIndexRowProps) {
             return <VirtualTableRowCell key={name} className={styles[name]} />;
           }
 
-          const { episodeCount: seasonEpisodeCount = 0, episodeFileCount: seasonEpisodeFileCount = 0, totalEpisodeCount: seasonTotalEpisodeCount = 0 } = latestSeason.statistics || {};
+          const { episodeCount: seasonEpisodeCount = 0, episodeFileCount: seasonEpisodeFileCount = 0, totalEpisodeCount: seasonTotalEpisodeCount = 0 } = latestSeason.statistics ?? {};
 
           return (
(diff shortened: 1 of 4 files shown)


frontend/src/Series/Index/Table/SeriesIndexRow.tsx
```diff
@@ -324,7 +324,7 @@ function SeriesIndexRow(props: SeriesIndexRowProps) {
             return <VirtualTableRowCell key={name} className={styles[name]} />;
           }
 
-          const seasonStatistics = latestSeason.statistics || {};
+          const { episodeCount: seasonEpisodeCount = 0, episodeFileCount: seasonEpisodeFileCount = 0, totalEpisodeCount: seasonTotalEpisodeCount = 0 } = latestSeason.statistics || {} as Statistics;
```
Gykes: `{} as Statistics` is an unsafe cast, the empty object doesn't actually satisfy Statistics. You could use the same `?? {}` pattern from your `DeleteSeriesModalContent` fix with defaults

we should probably also make `Series.statistics` in `Series.ts` optional as well. If an unrefreshed site is missing `Season.statistics` it will likely cause a similar error. Chances are there are more issues like this lurking around as well

Gykes (commented):
Static code review only. Thanks for taking care of this. Only found 1 small issue and a potential addition to solve something in the future. 

Also please run `yarn lint --fix` to fix the prettier issues.

frontend/src/Series/Index/Table/SeriesIndexRow.tsx
Venipa: yes thanks, I've corrected it, it was indeed not necessary

Venipa: I've only encountered the statistics undefined with the season model so far

Gykes (approved):

PR #1039, opened 2026-03-01
Fix: Downloads stuck in "Importing" due to external ID parsing false positives + Notification error/Crashing

### LINKED ISSUE #1038: [V2] Torrents fail to import after download on latest nightly

### Current Behavior

As mentioned on discord, the latest nightly no longer always imports new downloads until the windows service is restarted. I've verified the service account for the whisparr service has full access to read/write/delete permissions. Previous 2.0 builds worked fine. 

### Expected Behavior

Downloads typically import automatically. 

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: Windows
- Whisparr: V2
- Docker Install: No
- Using Reverse Proxy: No
- Browser: Any
- Database:
```

Develop (V2)

### Trace Logs?

[whisparr.trace (1).txt](https://github.com/user-attachments/files/25650536/whisparr.trace.1.txt)

### Anything else?

_No response_

Gykes:
Downloads with titles like `Hands-On [01+22] - Mario Gets A Very Hands-On Massage` get permanently stuck in "Downloaded - Importing" with no manual import icon. The series name "Hands-On" is falsely parsed as a JAV catalog number, `FindByExternalId("hands-on")` finds multiple episodes, and `SingleOrDefault()` throws, crashing the download monitoring loop on every pass.

I went ahead and fixed `FindByExternalId` to return null when multiple episodes match the same external ID, forcing manual import instead of crashing with `InvalidOperationException`. I also tightned up the external ID regex to help prevent further bad data.

Fixed `NotificationService` crash when processing `ManualInteractionRequiredEvent` with an empty episode list

fixes: https://github.com/Whisparr/Whisparr/issues/1038

### DIFF
--- a/src/NzbDrone.Core/Notifications/NotificationService.cs
+++ b/src/NzbDrone.Core/Notifications/NotificationService.cs
@@ -217,7 +217,7 @@ namespace NzbDrone.Core.Notifications
             var remoteEpisode = message.Episode;
             var series = remoteEpisode?.Series;
 
-if (remoteEpisode != null && remoteEpisode.Episodes?.Any() != true)
+            if (remoteEpisode != null && remoteEpisode.Episodes?.Any() != true)
             {
                 _logger.Trace("Manual interaction required for {0}, but no episode information is available", message.TrackedDownload.DownloadItem.Title);
             }
--- a/src/NzbDrone.Core/Notifications/NotificationService.cs
+++ b/src/NzbDrone.Core/Notifications/NotificationService.cs
@@ -217,7 +217,7 @@ namespace NzbDrone.Core.Notifications
             var remoteEpisode = message.Episode;
             var series = remoteEpisode?.Series;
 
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Notifications/NotificationService.cs
```diff
@@ -217,9 +217,14 @@ public void Handle(ManualInteractionRequiredEvent message)
             var remoteEpisode = message.Episode;
             var series = remoteEpisode?.Series;
 
+            if (remoteEpisode != null && (remoteEpisode.Episodes == null || !remoteEpisode.Episodes.Any()))
```
plz12345:
```suggestion
if (remoteEpisode != null && remoteEpisode.Episodes?.Any() != true)
```

plz12345 (changes requested):
Just a small complexity reduction tweak.

plz12345 (dismissed):

PR #1085, opened 2026-05-19
Chore: Bump MailKit and System.Text.Encoding.CodePages

Gykes:
updating dependencies


Gykes:
@plz12345 Should be good now. I tried it update to 4.16 but they dropped .net6 support so went back to 4.8 and just accepted the negative.

PR #1090, opened 2026-05-22
Fix: Fix Scene History

### LINKED ISSUE #1089: Episode history fields mismatch

### Current Behavior

Fields date, details and custom format score are not correctly displayed in episode history window  

Image

### Expected Behavior

Fields are in correct order

### Steps To Reproduce

Open episode history tab

### Environment

```markdown
- OS: Ubuntu
- Whisparr: 2.2.deveolp
- Docker Install: No
- Using Reverse Proxy: No
- Browser: Chrome
- Database:
```

V2

### Trace Logs?

No logs.

### Anything else?

_No response_

Gykes:
At some point extra table cell rows were created causing most of them to shift over out of alignment. 

Closes: https://github.com/Whisparr/Whisparr/issues/1089

Fixed UI: 
image

### DIFF
--- a/frontend/src/Episode/History/EpisodeHistoryRow.js
+++ b/frontend/src/Episode/History/EpisodeHistoryRow.js
@@ -102,14 +102,6 @@ class EpisodeHistoryRow extends Component {
           />
         </TableRowCell>
 
-        <TableRowCell>
-          <EpisodeFormats formats={customFormats} />
-        </TableRowCell>
-
-        <TableRowCell>
-          {formatCustomFormatScore(customFormatScore, customFormats.length)}
-        </TableRowCell>
-
         <RelativeDateCell
           date={date}
           includeSeconds={true}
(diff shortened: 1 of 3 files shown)


frontend/src/Episode/History/EpisodeHistoryRow.js
plz12345, line 1: @Gykes 
I'm lost I think.  Why are you adding CSS for custom format, but also removing the fields from the row?

Gykes: The row had duplicate cells rendering `customFormat` data. This was causing duplicated data and for them to shift and go into the wrong columns. The row already has a consolidated cell (the tooltip one slightly further down the page). 

The Tooltip version has `className={styles.customFormatScore}` in the JSX but it was never defined. I just made it 55px to pretty much match other of similar style like `HistoryRow.css`.

plz12345 (approved):

PR #1099, opened 2026-06-08
Fix: Replace SonarrBranding with Whisparr Across UI Links

Mothball7205:
#### Database Migration
NO

#### Description
update system info links from Sonarr to Whisparr in the System status page, file browser modal.
Also removed reference to sonarr wiki for authentication. I couldn't find a whisparr equivalent.

let me know if this is fine or if some changes are needed

image
Screenshot From 2026-06-08 19-29-02


Gykes (approved):
Thanks for this, looks good

Issue #1300, opened 2026-09-18
OpenAPI: the served document does not describe the V3 API


alextomas955:
`src/Whisparr.Api.V3/openapi.json` is committed and is meant to be the client-generation
artifact for the V3 API. Pointed at a generator today it produces a client that cannot read
most responses, cannot name its own methods stably, and cannot handle any failure.

Everything below was checked against a Debug instance built from `v2-develop` at edb091cf7,
started against an empty data directory, with the document fetched from
`http://localhost:6969/docs/v3/openapi.json` rather than read from source.

### The committed file is stale, and `docs.sh` cannot regenerate it

`docs.sh` does not run. Two independent reasons:

1. It publishes to and reads from `$outputFolder/net8.0/$RUNTIME/`. Every project targets
   **net10.0**.
2. It pins `Swashbuckle.AspNetCore.Cli` **6.3.0**, a net6.0 tool that cannot load a net10.0
   assembly:

   ```
   Framework: 'Microsoft.AspNetCore.App', version '6.0.0-rc.2.21480.10' -> not found
   ```

The repo references `Swashbuckle.AspNetCore.SwaggerGen 10.2.3`. With the CLI at 10.2.3 it
works, and its output is byte for byte identical to what the running server serves, so the
tool and the server agree once the version does.

The committed file is `openapi: 3.0.1` with 228 operations. The server serves `3.0.4` with
230. `/api/v3/customformat/bulk`, `/api/v3/importlistexclusion/paged`, `/api/v3/indexerflag`
and `/api/v3/rename/bulk` exist and are undocumented; `/api/v3/config/importlist` is
documented and gone.

### What the document gets wrong

**No `operationId`, on any operation.** 230 of 230. Generators invent method names from the
URL, so every client picks different names and they churn whenever a path changes.

**No 4xx or 5xx, on any operation.** `200` is the only key that appears under `responses`
anywhere in the document. Nothing can handle failure in a typed way.

**80 operations declare `200` with no content.** Not all of these are wrong: 16 are the
`[RestDeleteById]` actions, which return `void` and really do answer an empty 200
(`DELETE /api/v3/tag/1` -> 200, no `Content-Type`, 0 bytes). The rest return real data through
`object` or `IActionResult` and are simply undescribed.

**Actions that answer `201` or `202` are documented as `200`.** Captured:

| | declared | actual |
|---|---|---|
| `POST /api/v3/command` | 200 | **201** |
[...]

PR #1302, opened 2026-09-18
Chore: Repair docs.sh and regenerate openapi.json

alextomas955:
#### Database Migration

NO

#### Description

`docs.sh` cannot run. It publishes to and reads from `$outputFolder/net8.0/$RUNTIME/` while
every project targets **net10.0**, and it pins `Swashbuckle.AspNetCore.Cli` **6.3.0**, a net6.0
tool that cannot load a net10.0 assembly:

```
Framework: 'Microsoft.AspNetCore.App', version '6.0.0-rc.2.21480.10' -> not found
```

Three further faults meant it could not have worked on Linux or macOS even with the framework
resolved:

- the Linux and Mac arms assign `WHERE` instead of `RUNTIME`, so the runtime identifier is empty
- the assembly is spelled `whisparr.console.dll`, which only resolves on a case-insensitive
  filesystem
- `dotnet new tool-manifest` fails when a manifest already exists

The CLI version is now read from `Whisparr.Host.csproj` rather than written a second time, so it
cannot drift from `Swashbuckle.AspNetCore.SwaggerGen` again. The target framework is matched by
glob for the same reason. Those two hand-copied facts are what broke.

I ran the repaired script end to end on Windows. The document it produces is **byte for byte
identical** to what a Debug instance serves at `/docs/v3/openapi.json`, so the tool and the server
agree once the version does.

The .NET 10 SDK writes the tool manifest to the repository root rather than to `.config`, which
`.gitignore` already covers, so the root path is ignored too.

#### What the regeneration changes

The committed document was last generated a long way back. Regenerating moves it from
`openapi 3.0.1` to `3.0.4` and from 228 operations to 230:

| | |
|---|---|
| Paths that exist but were undocumented | `/api/v3/customformat/bulk`, `/api/v3/importlistexclusion/paged`, `/api/v3/indexerflag`, `/api/v3/rename/bulk` |
| Paths documented but gone | `/api/v3/config/importlist`, `/api/v3/config/importlist/{id}` |
| Schemas removed | `DateOnly`, `DayOfWeek`, `ImportListConfigResource`, `ListSyncLevelType`, `Rejection`, `TimeSpan`, `Version` |
| Schemas added | `CustomFormatBulkResource`, `ImportListExclusionResourcePagingResource`, `ImportRejectionResource`, `IndexerFlagResource`, `NewItemMonitorTypes`, `QueueStatus`, `SeriesTypes` |

No source behaviour changes. This is the baseline the other pull requests on #1300 are measured
against, and it is deliberately on its own so their diffs are readable.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The check is that the script runs and that its output matches the running server
byte for byte. `Whisparr.Api.Test` passes 34/34.


Gykes:
I'm currently in the process of fixing the issue.

alextomas955:
Superseded by #1301 (60766ac), which lands a better answer to the same problem: `openapi.json` is no longer committed at all, CI generates it from a running instance and attaches it to each release, and `/docs` now serves the document and a UI in Release builds rather than only Debug.

Nothing here survives that. This PR repaired `docs.sh` and regenerated the committed document; both files are now deleted. The `.gitignore` entry it added was only needed because `docs.sh` created a local tool manifest, so that goes too.

For the record, since it was measured rather than guessed: the repaired script did run, and its output was byte for byte identical to what a Debug instance served at `/docs/v3/openapi.json`. The regeneration moved the document from `openapi 3.0.1` to `3.0.4` and from 228 operations to 230, with `/api/v3/customformat/bulk`, `/api/v3/importlistexclusion/paged`, `/api/v3/indexerflag` and `/api/v3/rename/bulk` appearing and `/api/v3/config/importlist` disappearing. That drift is what #1301 makes impossible from here on.

The other pull requests on #1300 are unaffected in substance and have been rebased onto 60766ac.

PR #1305, opened 2026-09-18
Chore: Document the six actions that answer 202 Accepted

alextomas955:
#### Database Migration

NO

#### Description

Six actions call `Accepted(...)`, so the response has always been `202` with a body. The served
document declares `200` with no content, which is wrong about both the status and the payload.

Each now declares the type it returns and carries an attribute naming the real status:

```diff
+[ProducesResponseType(typeof(List<QualityDefinitionResource>), StatusCodes.Status202Accepted)]
-public object UpdateMany([FromBody] List<QualityDefinitionResource> resource)
+public ActionResult<List<QualityDefinitionResource>> UpdateMany([FromBody] List<QualityDefinitionResource> resource)
```

`SeasonPassController.UpdateAll` returns `Accepted(new object())`, an empty object body with no
shape worth describing, so it gets the status alone and keeps `IActionResult`. Empty `{}` bodies
are a separate question and are left as they are.

`ProducesResponseType` is new to this codebase. A status code cannot be expressed in a return type,
so it is the tool for this. `StatusCodes` is already used in `PingController`, and `PingController`
already returns `ActionResult<T>` while only ever returning a result object, so the shape of the
change follows what is here.

#### What was captured

All six, on a running instance built from this branch and from one built from `v2-develop`, both
starting from the same empty data directory seeded with the same single series, episode and episode
file so that ids line up. Status and body compared, all six identical:

| | | |
|---|---|---|
| `PUT /api/v3/episode/monitor` | 202 | 322 bytes |
| `PUT /api/v3/episodefile/editor` | 202 | 753 bytes |
| `PUT /api/v3/episodefile/bulk` | 202 | 788 bytes |
| `PUT /api/v3/qualitydefinition/update` | 202 | 4,080 bytes |
| `PUT /api/v3/series/editor` | 202 | 708 bytes |
| `POST /api/v3/seasonpass` | 202 | 2 bytes |

The two `episodefile` endpoints throw before reaching their return without existing files, which is
why the instances were seeded rather than left empty.

#### Worth being explicit about

The `200` is **replaced, not joined**. Swashbuckle drops the inferred `200` once an explicit
`ProducesResponseType` is present, so these operations now declare `202` alone. That is what the
server does, and it never answered `200` on any of these paths, but it is a visible change to the
document rather than an addition to it.

Exactly those six operations change, no schema is added or removed, and empty `200` responses in
the served document drop from 80 to 74.

Sonarr's v5 API moved the equivalent actions to `200` rather than documenting the `202`, for
example `Ok<List<QualityDefinitionResource>> UpdateMany`. Doing that here would change an existing
response code on a live API with no second version to break it in, so this describes what the API
does instead. If you would rather correct the status than describe it, that is a different change
and I am happy to do it that way.

The 35 remaining actions that answer `201` or `202` while the document says `200` already declare a
response type, so they need the status attribute only. They are #1306.

commit.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The status and body were captured on both sides rather than asserted.
`Whisparr.Api.Test` passes 34/34.


alextomas955:
Superseded by #1310, which documents all six of these as `202`.

Confirmed against a document generated from 7c6d179. All six carry the same status and the same schema this PR produced, byte for byte in the response object:

```
PUT  /api/v3/episode/monitor            202  array of EpisodeResource
PUT  /api/v3/episodefile/editor         202  array of EpisodeFileResource
PUT  /api/v3/episodefile/bulk           202  array of EpisodeFileResource
PUT  /api/v3/qualitydefinition/update   202  array of QualityDefinitionResource
PUT  /api/v3/series/editor              202  array of SeriesResource
POST /api/v3/seasonpass                 202  no schema
```

The only thing here that #1310 does not do is change the declared return type from `object` to `ActionResult<List<T>>`. #1310 names the type in the attribute instead and leaves the signature alone. That gives the same document; the compile-time guarantee is weaker, but not enough to justify a competing change to the same eight lines.

PR #1306, opened 2026-09-18
Chore: Document the 201 and 202 answers on the remaining create and update actions

alextomas955:
#### Database Migration

NO

#### Description

Thirty-five actions call `Created(...)` or `Accepted(...)` and already declare a response type, so
the document has the right body and the wrong status: every one of them says `200`. A generated
client that only handles `200` treats every create and every bulk edit as an error.

Each gains one attribute and nothing else:

```diff
+[ProducesResponseType(StatusCodes.Status201Created)]
 public ActionResult<TagResource> Create([FromBody] TagResource resource)
```

The attribute carries no type on purpose. Swashbuckle falls back to the action's return type for
the body when `ProducesResponseType` names only a status, so the schemas are untouched. It also has
to be type-less on `ProviderControllerBase`, where an attribute argument cannot use the class type
parameter.

Three of the thirty-five are on `ProviderControllerBase`, which is why thirty-five actions move
**forty-seven operations**: they are the create, update and bulk update of every download client,
import list, indexer, metadata and notification provider. `ConfigController` accounts for another
six the same way.

#### What was captured, and what was not

Thirty-four of the thirty-five were driven on a running instance built from this branch and from
one built from `v2-develop`, both starting from the same empty data directory seeded with the same
single series, episode and episode file.

- **32 identical, byte for byte**
- **4 same status, bodies differing only in values that cannot repeat across two runs**

| | |
|---|---|
| `POST /api/v3/command` | `queued`, `started` and `stateChangeTime` timestamps |
| `PUT /api/v3/config/host` | the per-instance `apiKey` |
| `POST /api/v3/rootfolder` | `freeSpace` |
| `POST /api/v3/series` | the `traceId` of a 400 that both builds return |

Seventeen answered `201` and seventeen answered `202` across those runs, matching what the
attributes now declare.

The one that was **not** driven is `SeriesController.AddSeries`. It needs a metadata lookup to
reach its return, and the 400 it gives without one is the same on both builds. It is
`Created(series.Id)` on the only non-throwing path, in a `RestController` whose `Created(int)`
helper returns `CreatedAtAction` unconditionally, so `201` is not conditional on anything an
instance could vary.

#### Two pre-existing problems this does not touch

Both found while checking the above, both worth their own change:

- `PUT` on `customformat/bulk`, `downloadclient/bulk`, `importlist/bulk` and `indexer/bulk` returns
  an **array** while the declared type is a single resource. The status is now right and the body
  schema stays wrong.
- `PUT /api/v3/remotepathmapping/{id}` returns the domain model rather than the resource. The two
  happen to serialise to the same four fields, so the document is accidentally correct.

#### Worth being explicit about

The `200` is **replaced, not joined**. Swashbuckle drops the inferred `200` once an explicit
`ProducesResponseType` is present. The server never answered `200` on any of these paths, so this
is a correction, but it is a visible change to the document rather than an addition to it.

Forty-seven operations change status, seventeen to `201` and thirty to `202`. No schema is added,
removed or altered, and the count of empty `200` responses is unchanged at 80.

commit.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The change is an attribute per action and the check is the live capture above.
`Whisparr.Api.Test` passes 34/34.

Follows #1305, which covers the six actions that also needed a declared type.


alextomas955:
Superseded by #1310, and this approach was worse. Both worth stating.

#1310 reaches the same 17 `201` and 36 `202` operations, verified against a document generated
from 7c6d179.

It gets there a better way. `SuccessStatusCodeOperationFilter` **moves** the existing `200`
response object to the new status code, so the schema inferred from `ActionResult<T>` travels with
it. This PR added a type-less `[ProducesResponseType(StatusCodes.Status202Accepted)]` and claimed
that Swashbuckle would fall back to the return type for the body.

That claim is wrong, and I should correct it rather than leave it in a closed PR. It holds for the
`201` operations but not the `202` ones. Measured on this branch against its base, **30 operations
lost their response schema**, every `202` in the set:

```
POST /api/v3/languageprofile            PUT /api/v3/importlist/{id}
PUT  /api/v3/autotagging/{id}           PUT /api/v3/importlistexclusion/{id}
PUT  /api/v3/config/downloadclient/{id} PUT /api/v3/indexer/bulk
PUT  /api/v3/config/host/{id}           PUT /api/v3/indexer/{id}
PUT  /api/v3/config/indexer/{id}        PUT /api/v3/languageprofile/{id}
PUT  /api/v3/config/mediamanagement/{id} PUT /api/v3/metadata/{id}
PUT  /api/v3/config/naming/{id}         PUT /api/v3/notification/{id}
PUT  /api/v3/config/ui/{id}             PUT /api/v3/qualitydefinition/{id}
PUT  /api/v3/customfilter/{id}          PUT /api/v3/qualityprofile/{id}
PUT  /api/v3/customformat/bulk          PUT /api/v3/releaseprofile/{id}
PUT  /api/v3/customformat/{id}          PUT /api/v3/remotepathmapping/{id}
PUT  /api/v3/delayprofile/{id}          PUT /api/v3/series/{id}
PUT  /api/v3/downloadclient/bulk        PUT /api/v3/tag/{id}
PUT  /api/v3/downloadclient/{id}        PUT /api/v3/episode/{id}
PUT  /api/v3/importlist/bulk            PUT /api/v3/episodefile/{id}
```

My check missed it because I counted `200` responses with no content, which stayed at 80. These
became empty **202s**, which that count does not see. Measuring the thing I had named rather than
the thing I cared about is what hid it.

#1310 keeps the schema on all of them. `PUT /api/v3/tag/{id}` is `202` with `TagResource` there and
`202` with nothing here. It also does two things this PR explicitly declined to: it declares the
list type on `PUT /api/v3/customformat/bulk`, and it drops the wrong single-resource schema on the
provider bulk endpoints instead of leaving it in place at a corrected status.

Nothing here is worth salvaging.

PR #1307, opened 2026-09-18
Chore: Stop documenting anonymous endpoints as needing an API key

alextomas955:
#### Database Migration

NO

#### Description

`AddSwaggerGen` registers a document level security requirement for both the `X-Api-Key` header and
the `apikey` query parameter, and nothing overrides it per operation. Six operations serve
anonymous traffic and are documented as rejecting it, `GET /ping` among them.

An `IOperationFilter` now gives any operation whose endpoint metadata carries `IAllowAnonymous` an
empty security list, which is how OpenAPI expresses "this one does not need what the document
asks for". The document level requirement is untouched, so every other operation is unchanged.

Reading `IAllowAnonymous` off `EndpointMetadata` rather than looking for the attribute on the method
picks up controller level `[AllowAnonymous]` as well, which is how `AuthenticationController`
declares it.

#### What was captured

Exactly six operations change, and nothing else in the document does. All six were called on a
running instance with and without a key. They answer the same either way, and none of them answers
`401`:

| | no key | with key | |
|---|---|---|---|
| `GET /ping` | 200 | 200 | `PingController.GetStatus` |
| `HEAD /ping` | 200 | 200 | `PingController.GetStatus` |
| `GET /login` | 404 | 404 | `StaticResourceController.LoginPage` |
| `POST /login` | 302 | 302 | `AuthenticationController.Login` |
| `GET /logout` | 302 | 302 | `AuthenticationController.Logout` |
| `GET /content/{path}` | 404 | 404 | `StaticResourceController.IndexContent` |

The two 404s are the SPA files, which are not present in a backend-only Debug output. What matters
is that they are not 401. Four protected endpoints sampled alongside them still answer `401`
without a key and `200` with one: `/api`, `/api/v3/system/status`, `/api/v3/tag` and
`/api/v3/series`.

Note that `GET /` and `GET /{path}` are deliberately **not** in the list.
`StaticResourceController.Index` has no `[AllowAnonymous]`, so the API key requirement on those is
what the server does.

This changes the document only. No middleware, policy or attribute is touched.

commit.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. A test that asserted which operations carry an empty security list would restate the
filter rather than check it; the live capture above is what shows the document now matches the
server. `Whisparr.Api.Test` passes 34/34.


alextomas955:
Superseded by #1310, which does the same thing with `AllowAnonymousOperationFilter`.

Confirmed against a document generated from 7c6d179: `GET /ping`, `HEAD /ping`, `POST /login` and `GET /logout` now carry `security: []`, and the document level requirement is untouched. That is four rather than the six here because #1310 also excludes `StaticResourceController` from the document, so `GET /login` and `GET /content/{path}` no longer appear at all. Same outcome by a different route, and it removes the malformed `paths["/"]` at the same time, which this PR did not.

PR #1308, opened 2026-09-18
Chore: Stop the command schema forbidding the arguments commands require

alextomas955:
#### Database Migration

NO

#### Description

`CommandResource` is documented with `additionalProperties: false`, and the server requires
properties that are not on it. A client generated from the document cannot start any command that
takes arguments.

`CommandController.StartCommand` seeks the request body back to zero and deserialises the raw JSON
into the concrete `Command` subtype rather than using the bound `CommandResource`. `Startup`
enables `BufferingMiddleware` for `/api/v3/command` specifically so it can. The arguments a command
needs therefore live on the `Command` type and never appear on the resource.

A schema filter now allows additional properties on `CommandResource` alone.

#### What was captured

Against a running instance built from this branch:

```
POST /api/v3/command  {"name":"RefreshSeries","seriesId":1}
-> 201, body.seriesIds == [1]
```

That exact body validated against the `CommandResource` schema from each document:

| | |
|---|---|
| before | **INVALID** `Additional properties are not allowed ('seriesId' was unexpected)` |
| after | **VALID** |

So the server accepts a request that its own document forbids, and that is what this closes.

#### Why only this one schema

`additionalProperties: false` is a Swashbuckle default that lands on all 98 object schemas in this
document. The other 97 are left alone. For every other resource the server binds the body to the
resource and `System.Text.Json` ignores what it does not recognise, so a strict schema there
describes the contract rather than contradicting it. Command is the one place where the strictness
is wrong.

One schema changes, no operation changes, and no response body changes.

commit.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The check is the schema validation above, run against both documents with the body
the server actually accepts. `Whisparr.Api.Test` passes 34/34.


alextomas955:
Superseded by #1310, which lands a `CommandResourceSchemaFilter` doing the same thing under the same name.

Confirmed against a document generated from 7c6d179: `CommandResource` no longer carries `additionalProperties: false`, and the other 97 object schemas keep Swashbuckle's default. That is the same scope this PR argued for.

PR #1309, opened 2026-09-18
Chore: Give every operation a stable operationId

alextomas955:
#### Database Migration

NO

#### Description

Not one of the 231 operations has an `operationId`. Generators fall back to inventing method names
from the URL, so every client picks different names and they churn whenever a path changes.

Operations are now named after the controller and action that serve them, which is what does not
change when a route does:

```
AutoTagging_GetAll        GET  /api/v3/autotagging
Backup_Restore            POST /api/v3/system/backup/restore/{id}
Tag_Create                POST /api/v3/tag
```

#### The nine collisions

Nine of those names are not unique, because an action can serve several routes or several methods.
A document filter breaks each collision with the last path segment, and where that is still not
enough, with the HTTP method:

| operationId | | |
|---|---|---|
| `CustomFormat_UpdateById` | PUT | `/api/v3/customformat/{id}` |
| `CustomFormat_UpdateBulk` | PUT | `/api/v3/customformat/bulk` |
| `DownloadClient_UpdateProviderById` | PUT | `/api/v3/downloadclient/{id}` |
| `DownloadClient_UpdateProviderBulk` | PUT | `/api/v3/downloadclient/bulk` |
| `EpisodeFile_SetQualityById` | PUT | `/api/v3/episodefile/{id}` |
| `EpisodeFile_SetQualityEditor` | PUT | `/api/v3/episodefile/editor` |
| `ImportList_UpdateProviderById` | PUT | `/api/v3/importlist/{id}` |
| `ImportList_UpdateProviderBulk` | PUT | `/api/v3/importlist/bulk` |
| `Indexer_UpdateProviderById` | PUT | `/api/v3/indexer/{id}` |
| `Indexer_UpdateProviderBulk` | PUT | `/api/v3/indexer/bulk` |
| `QueueAction_GrabById` | POST | `/api/v3/queue/grab/{id}` |
| `QueueAction_GrabBulk` | POST | `/api/v3/queue/grab/bulk` |
| `RenameEpisode_GetEpisodes` | GET | `/api/v3/rename` |
| `RenameEpisode_GetEpisodesBulk` | GET | `/api/v3/rename/bulk` |
| `StaticResource_Index` | GET | `/` |
| `StaticResource_IndexByPath` | GET | `/{path}` |
| `Ping_GetStatusGet` | GET | `/ping` |
| `Ping_GetStatusHead` | HEAD | `/ping` |

A literal tail that is the controller's own root segment is dropped, which is why `/api/v3/rename`
keeps the plain name and only `/api/v3/rename/bulk` is suffixed.

The filter groups by the colliding name and decides from the group, so the result does not depend
on the order operations are visited, and an id only changes when a colliding sibling is added or
removed.

#### What was captured

On a running instance built from this branch: **231 operations, 231 ids, no duplicates**. The
generated document is identical to the one from `v2-develop` once `operationId` is removed from
both, so nothing else moved.

#### Precedent

None to follow, and I would rather say so than cite loosely. Sonarr configures no operation ids in
any of its API projects, including the v5 rewrite, so their document has the same gap. The scheme
here is Swashbuckle's own `controller_action` convention with the collisions resolved.

`openapi.json` is generated by `docs.sh` and is not edited here. That script does not currently run
and is repaired in #1302, so the committed document is regenerated there.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. A test asserting the ids would restate the filter; what matters is that the generated
document has 231 unique ids and is otherwise unchanged, which was checked against a running
instance. `Whisparr.Api.Test` passes 34/34.


alextomas955:
Closing this one, not wanted.

PR #1316, opened 2026-09-19
Chore: Describe the responses the V3 API actually sends

### LINKED ISSUE #1315: OpenAPI: the document does not describe the responses the V3 API sends

### Current Behavior

The OpenAPI document served at `/docs/v3/openapi.json` describes responses the V3 API does not send, and omits responses it does. Forty-four operations return data the document says nothing about, one operation's response contradicts its own schema, one endpoint's request body is undocumented so the document gives no way to call it, and six status codes are not the ones returned.

#1310 fixed the operation ids, the REST convention status codes, the anonymous endpoint security and the command schema. What follows is what is left, measured against a Debug instance built from 536c7879 with the document fetched from the running server rather than read from source.

**Seventy-four operations declare a success with no content.** Thirty are `[RestDeleteById]` actions that return `void` and really do answer an empty 200, so those are correct. The other forty-four return data:

| | |
|---|---|
| return `object` or `IActionResult` | `ApiInfoController.GetApiInfo`, `AutoTaggingController.GetTemplates`, `CustomFormatController.GetTemplates`, `NamingConfigController.GetExamples`, `FileSystemController.GetContents`, `SystemController.DuplicateRoutes`, `SeriesImportController.Import`, `ManualImportController.ReprocessItems`, `ReleaseController.DownloadRelease`, `ProviderControllerBase.TestAll` |
| return an anonymous type | `filesystem/type`, `filesystem/mediafiles`, `system/shutdown`, `system/restart`, both backup restores |
| return a file with no media type declared | calendar feed, both log file reads, media covers |
| undescribed on all five provider types | `POST /*/test`, `POST /*/action/{name}` |

`GET /api` is the plainest case: the action builds an `ApiInfoResource` and returns it as `object`. `ProviderControllerBase` is generic, so each of its three undescribed actions covers five operations.

**One schema contradicts the wire.** `HealthResource.wikiUrl` is documented as an object with `fullUri`, `scheme`, `host`, `port`, `path`, `query` and `fragment`. A string is sent. `STJHttpUriConverter` writes `value.FullUri`, so the structural description of `HttpUri` has never matched. It is the only operation in the document whose 200 response fails validation against its own declared schema. `Version`, `TimeSpan` and `DateOnly` have converters too and are already described as strings, so this is the last one.

**Seventy-seven request bodies are mandatory and documented as optional.** Nullable reference types are off in this tree, so ApiExplorer reports every body as not required. No `[FromBody]` parameter in the API is nullable or carries a default, and a request without one is rejected before the action runs.

**One endpoint's request body is undocumented entirely.** `POST /api/v3/system/backup/restore/upload` requires a multipart file and declares no request body at all, because `UploadAndRestore` reads `Request.Form.Files` directly instead of binding a parameter. Nothing in the document says a file has to be sent, or under what field name.

**Six status codes are wrong.**

| | declared | actual |
|---|---|---|
| `POST /login` | 200 | 302, or 401 when the data protection key ring is corrupt |
| `GET /logout` | 200 | 302 |
| `GET /api/v3/parse` with no title | 200 | 204 |
| `GET /api/v3/system/task/{id}` with an unknown id | 200 | 204 |
| `GET /feed/v3/calendar/whisparr.ics` when empty | 200 | 204 |
| `POST /*/testall` with a failing provider | 200 only | also 400, carrying the same list |

Both 204s come from an action returning `null`, which MVC serialises as 204. `TaskController` is the surprising one: it overrides `GetResourceById` to return `null` rather than letting `ModelNotFoundException` reach `RestController`, so it alone among the by-id reads answers 204 where every other one answers 404.

**One content type is simply wrong.** `GET /api/v3/system/routes` carries `[Produces("application/json")]` and returns `Content(graph, "text/plain")`, a DOT graph of the route table.

### Expected Behavior

The document describes what the server sends, so that reading it tells you what an endpoint returns and what it has to be given.

Concretely: every operation that returns data declares a schema for it; `HealthResource.wikiUrl` is a string; mandatory request bodies are marked required; the backup upload declares its multipart body; the six status codes above are the ones declared; and `/api/v3/system/routes` declares `text/plain`.

Nothing on the wire should move. V2 has no second API version to break things in, so widening a declared type or adding a status code to the document is in scope, and changing a status code, a property name, or an empty `{}` body to a 204 is not.

### Steps To Reproduce

1. Build and run from 536c7879, then fetch `http://localhost:6969/docs/v3/openapi.json`.
2. Count the operations declaring a success with no content: 74.
3. Compare a few against the server:

```
curl -H "X-Api-Key: $KEY" localhost:6969/api/v3/health
  -> wikiUrl is "https://wiki.servarr.com/..."   the schema says it is an object

curl localhost:6969/api/v3/parse
  -> 204, 0 bytes                                the document declares 200 only

curl -o /dev/null -w '%{http_code}' -X POST -H "X-Api-Key: $KEY" \
     -H "Content-Type: application/json" localhost:6969/api/v3/tag
  -> 400 on an absent body                       the document says the body is optional
```

4. Or validate a live response against the schema the document declares for it. `GET /api/v3/health` is the one that fails: `wikiUrl` arrives as a string and the schema says it is an object of seven properties. It is the only operation in the document where that check fails.

### Environment

- OS: Windows 11 (10.0.26200)
- Whisparr: v2-develop at 536c7879, local Debug build (reports 10.0.0.25521)
- Docker Install: No
- Using Reverse Proxy: No
- Browser: Not applicable, this is not a UI issue
- Database: SQLite 3.53.4

V2 (develop)

### Trace Logs

<details><summary>click to expand</summary>

Trace logs do not show this. Nothing here is a runtime fault, so the log is silent in every case below. The defect is the difference between what the server sends and what the document claims, and the evidence is that comparison, taken from a running instance built from 536c7879.

```
$ curl -s -o /dev/null -w '%{http_code} %{content_type}\n' -H "X-Api-Key: $KEY" \
       localhost:6969/api/v3/system/routes
200 text/plain
    document: [Produces("application/json")], and 200 with no content declared

$ curl -s -H "X-Api-Key: $KEY" localhost:6969/api/v3/health | head -c 200
[{"source":"IndexerRssCheck","type":"error","message":"No indexers available ...",
  "wikiUrl":"https://wiki.servarr.com/whisparr/system#no-indexers-available-..."}]
    document: wikiUrl -> $ref HttpUri, an object of seven properties

$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' -H "X-Api-Key: $KEY" \
       localhost:6969/api/v3/parse
204 0
$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' -H "X-Api-Key: $KEY" \
       localhost:6969/api/v3/system/task/999999
204 0
$ curl -s -o /dev/null -w '%{http_code}\n' -H "X-Api-Key: $KEY" \
       localhost:6969/api/v3/tag/999999
404
    the last is the contrast: every other by-id read answers 404, task answers 204
    document: 200 only, on all three

$ curl -s -o /dev/null -w '%{http_code}\n' -X POST -H "X-Api-Key: $KEY" \
       -H "Content-Type: application/json" localhost:6969/api/v3/tag
400
$ curl -s -o /dev/null -w '%{http_code}\n' -X POST localhost:6969/login
302
    document: both request bodies optional. Login's genuinely is, tag's is not

$ curl -s -o /dev/null -w '%{http_code} %{content_type}\n' -X POST -H "X-Api-Key: $KEY" \
       -H "Content-Type: application/json" -d "$METADATA" localhost:6969/api/v3/metadata/test
200 text/plain; charset=utf-8
    body is the two bytes 7b7d, that is {}
    document: 200 with no content declared

$ openapi-spec-validator openapi.json
    0 errors. The document is valid OpenAPI 3.0.4. Every defect above is a true statement
    about a document that is structurally well formed, which is why none of it surfaces as
    a validation failure or a log line.
```

</details>

### Anything else?

**Deliberately out of scope**, so this does not stall on decisions it does not need:

- **Error responses.** `200`, `201` and `202` are the only codes in the whole document. The API has three incompatible error shapes: `{message, description}` from `NzbDroneClientException`, RFC 9110 `application/problem+json` from the `[ApiController]` convention, and a bare array of validation failures from provider `test`. Documenting them needs a decision about which shape each operation claims, and guessing is worse than leaving it undocumented. The `400` on `testall` above is the exception, because it carries a domain type rather than one of those three.
- **Actions returning `new { }`.** Around fifteen operations answer with an empty object. There is no non-breaking fix: turning `{}` into a `204` would change the wire.
- **The provider bulk `PUT` schema.** `downloadclient/bulk`, `importlist/bulk` and `indexer/bulk` answer `202` with an array and are documented as `202` with no body. `customformat/bulk` is not affected: it is a concrete controller and #1310 typed it.

  The three that remain share `ProviderControllerBase`, so the response type is a type parameter and `[ProducesResponseType(typeof(List<TProviderResource>), 202)]` will not compile — `CS0416: an attribute argument cannot use type parameters`. A fix does exist: change the declared return type to `ActionResult<List<TProviderResource>>`, which is legal and which Whisparr-Eros has already shipped, drop the type-less attribute, and extend `SuccessStatusCodeOperationFilter` to cover the action so the inferred `200` moves to `202` with its schema intact. That is a change to a base class five controllers inherit plus the shared status filter, which is why it is listed here rather than folded in.

  For reference, none of the related projects has this fully right. Sonarr and Radarr declare `200` with a single resource, which is wrong about both the status and the shape. Eros fixed the shape and left the status at `200`. Sonarr V5 is correct only because the rewrite dropped the generic base and moved the endpoint to `200`, a wire change not available here.

**Not a defect, recorded so it is not re-reported.** The root security requirement declares both `X-Api-Key` and `apikey`. That is correct: the server supports both, and `[{A},{B}]` is the OpenAPI spelling of "A or B". Narrowing it to the header alone would mis-describe an authentication method the server accepts.

alextomas955:
#### Database Migration
NO

#### Description

Makes the served OpenAPI document describe the responses the V3 API actually sends. Ten commits, one concern each. The document changes throughout; nothing on the wire does.

#1310 covered the operation ids, the REST convention status codes, the anonymous endpoints and the command schema. This is the remainder.

**Responses that were not described at all**

- Ten actions returning `object` or `IActionResult` now declare the type they already return. `ProviderControllerBase.TestAll` is generic, so its one change covers five operations, and `ApiInfoController.GetApiInfo` was building a full `ApiInfoResource` and returning it as `object`.
- Six actions returning an anonymous type now return a named resource: `FileSystemTypeResource`, `FileSystemMediaFileResource`, `SystemShutdownResource`, `SystemRestartResource` and `BackupRestoreResource`. The names and property shapes are taken from `Sonarr.Api.V5` rather than invented, so a future sync of that area is a no-op instead of a conflict. Only the names and shapes are taken: their controllers return `Ok<T>`, a Minimal API `HttpResults` idiom from the V5 rewrite that does not belong in MVC controllers, so these return the resource directly. They deliberately do not derive from `RestResource`, which would have added an `id` that was never in these responses.
- The calendar feed, both log file reads and the media covers declare their media type and schema. `typeof(FileResult)` on the covers becomes `{"type": "string", "format": "binary"}` under each image type rather than a schema for the MVC result class, which was worth checking rather than assuming.
- `POST /*/test` and `POST /*/action/{name}` are described for all five provider types. `action/{name}` returns a payload that differs per provider, so `typeof(object)` is the only honest declaration.

**A schema that contradicted the wire**

`HealthResource.wikiUrl` was documented as an object with seven properties and is a string. `STJHttpUriConverter` writes `value.FullUri`, so `c.MapType<HttpUri>` now says so and the `HttpUri` schema drops out because nothing references it. It was the only operation whose 200 response failed validation against its own schema; it now passes. `Version`, `TimeSpan` and `DateOnly` have converters too and were already correct, so this was the last one.

**Request bodies**

An operation filter marks the body required wherever the action binds one from the body: 77 of 78. It keys on the binding source rather than on the presence of a request body, which is what correctly leaves `POST /login` alone, since that one binds `[FromForm]` and answers 302 on an empty body.

`POST /api/v3/system/backup/restore/upload` gains the multipart body it requires and had never declared. It is done with an operation filter rather than by binding an `IFormFile`: the action takes the first file whatever the field is called, and the UI sends `restore`, so naming a parameter `file` would have broken it.

**Status codes and content types**

`302` and `401` on login, `302` on logout, `204` on the calendar feed, on `GET /api/v3/parse` and on `GET /api/v3/system/task/{id}`, and the `400` that `testall` already returns with the same list type. `GET /api/v3/system/routes` had `[Produces("application/json")]` on an action returning `text/plain`.

The two 204s come from an action returning `null`, which MVC serialises as 204. `TaskController` overrides `GetResourceById` to return `null` rather than letting `ModelNotFoundException` reach `RestController`, so it alone among the by-id reads answers 204 where every other one answers 404.

#### Effect on the document

Measured from a running instance against one built from `v2-develop`:

| | before | after |
|---|---|---|
| operations | 227 | 227 |
| success with no content | 74 | **38** |
| request bodies marked required | 0 | **77** of 78 |
| schemas | 134 | 147 |
| response codes present | 200, 201, 202 | 200, 201, 202, 204, 302, 400, 401 |
| OpenAPI 3.0 validation errors | 0 | 0 |

102 operations change and every `operationId` stays unique. The 38 that remain are the 30 `[RestDeleteById]` actions that genuinely answer an empty 200, the `new { }` bodies, and the provider bulk `PUT`s, all of which are out of scope and noted in the issue.

A sample of what changes, so the shape of it is visible without reading the whole diff:

```
GET  /api                                     nothing -> ApiInfoResource
GET  /api/v3/filesystem/type                  nothing -> FileSystemTypeResource
GET  /api/v3/mediacover/{seriesId}/{filename} nothing -> image/*, binary
POST /api/v3/notification/testall             nothing -> array of ProviderTestAllResult
POST /api/v3/system/shutdown                  nothing -> SystemShutdownResource
POST /api/v3/system/backup/restore/upload     no body -> multipart, restore required
     HealthResource.wikiUrl                   object  -> string
```

#### Nothing on the wire changes

Every endpoint this touches was driven on a build of this branch and a build of `v2-develop`, and compared. Because `[Produces]` constrains content negotiation, the comparison varies the `Accept` header rather than relying on curl's default: **30 probes across four `Accept` values, 0 differ.**

That check earned its keep. An earlier revision of this branch put `[Produces("text/plain")]` on `ProviderControllerBase.Test`. That action returns `object` holding a string, so it is an `ObjectResult` and the constraint is real: a strict `Accept: application/json` got **406** instead of its body. Measured, reverted, and the endpoint now declares a string under all three formatter media types, which is what it actually negotiates. The web UI was never affected, since jQuery sends `*/*` in its Accept list and MVC ignores the header when it does, but a caller sending a strict `Accept: application/json` would have been.

The remaining `[Produces]` additions are on actions returning `ContentResult` or `PhysicalFileResult`, which ignore the attribute. The media covers were driven with real `.jpg`, `.png` and `.gif` files to confirm it, because `GetContentType` can fall back to `application/octet-stream`, which is not among the declared types.

None of the ten declared types has a subclass anywhere in the tree, so declaring the base cannot silently drop a derived type's properties.

Three things could not be driven and say so rather than claiming a check. `POST /api/v3/release` and `POST /api/v3/manualimport` each return their own `[FromBody]` parameter, so model binding constructs exactly the declared type and the two cannot diverge. `SeriesController.AddSeries` needs a metadata lookup to reach its return and is `Created(series.Id)` on its only non-throwing path.

Not a UI change.

#### Todos
- [x] Tests
- [ ] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)

One assertion added to `OpenApiFixture`: every operation declaring an `application/json` request body declares it required.

It is the only test here because it is the only change on the branch that lives as a single line of registration in `Startup.cs`, a file upstream syncs keep touching. Dropping that line in a merge takes all 77 bodies back to optional with a green build and nothing in the diff to notice. Everything else is an attribute on the action it describes, visible in any diff a reviewer is already reading.

Keying it on `application/json` rather than on the presence of a body is what lets it run without an exception list: the login form is multipart, so it falls out of the check rather than being excused from it.

Checked both ways. It passes, and with the registration removed it fails with

```
Expected boolean to be True because post /api/v3/autotagging binds a body, but found False
```

naming the operation rather than reporting a count.

`OpenApiFixture` passes 29/29 and `Whisparr.Api.Test` 34/34. The solution builds clean under `TreatWarningsAsErrors`.

No translation keys, this adds no user-facing strings. No wiki updates: the document is generated and published per release by #1301, so the wiki carries no copy of it.

* Fixes #1315


Gykes:
Various fixes were put in. I still have another branch that I need to test but will get to that later. The other branch should fix the error response shape as well as some other stuff.

